PR #2, opened 2023-04-10
LDAF-177 Add Nav component

### LINKED ISSUE #1: Add navigation to component library
Jira ticket: [LDAF-177](https://ldaf.atlassian.net/browse/LDAF-177)

## Background / problem statement

In order to build out a full navigation menu for both mobile and desktop, we'll need to support a two navigation component variations:
1. a navigation parent which will open a dropdown / subnav on hover, focus, and click (does not link anywhere)
2. a navigation item that appears in the main nav or the subnav (utilizing the `<Link/>` component)

At a stand-up on March 28 we discussed a variation of a navigation parent which would only open a subnav on hover and focus, and clicking would link to a page. The consensus was that this is a bit of a UX anti-pattern and we shouldn't support it, at least for now.

## Mocks

* [Header](https://www.figma.com/file/oGKbyCnCRRdNzLYbiags93/LDAF-Component-Library-USWDS-3.0.2?node-id=2196-12620&t=aGIFoQ1HdPeg5EG9-4)
* [Mobile Header](https://www.figma.com/file/oGKbyCnCRRdNzLYbiags93/LDAF-Component-Library-USWDS-3.0.2?node-id=2196-11223&t=aGIFoQ1HdPeg5EG9-4)

## Acceptance criteria

- [x] Support for a navigation item that links to a page and appears in the main nav menu.
- [x] Support for a navigation parent that opens a subnav menu (a standard dropdown on desktop and a collapsible panel on mobile).
- [x] Support for a navigation item that links to a page and appears in a subnav menu.

Since we're mostly focused on navigation elements that will appear in the header, we'll probably save any work on a possible sidenav ([mock in Figma](https://www.figma.com/file/oGKbyCnCRRdNzLYbiags93/LDAF-Component-Library-USWDS-3.0.2?node-id=2196-25149&t=aGIFoQ1HdPeg5EG9-4)) for later.

LouisFettet:
## Closes  #1

Jira ticket: [LDAF-177](https://ldaf.atlassian.net/browse/LDAF-177)

## Proposed changes

- adds a `Header` component that is mostly a shell for now; this was necessary to handle collapsing and expanding the `Nav` on mobile since USWDS is pretty opinionated about the layout (and there was no way to get all the elements we needed in one `Nav` component)
- adds a `Nav` component which can contain many `NavItem`s
- adds a `NavItem` component which supports both mouse and keyboard user interactions on both mobile and desktop and has 3 variations:
  1. basic nav link that appears in the main nav bar
  2. basic accordion nav menu which supports a single column of nav links
  3. slightly more complex accordion nav mega menu which supports a number of columns of nav links
- adds a new story for the `Header`, already [deployed to chromatic](https://641a2b7efd9419d13bbf8390-yugeewqjno.chromatic.com/?path=/story/example-header--default)
- adds a `util` dir with an implementation of `_.chunk` to make it easier to organize links into mega menus
  - implementation copied from [you-dont-need-lodash-underscore](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_chunk)
- moves `classNames` into the `util` dir and slightly refactors usage of the function so that it's a reactive value
  - this ensures that it gets properly captured in code coverage
  - there may be some side effects to this I'm unaware of; if this impacts performance in any way it may be better to use the previous approach
- adds a new `TestSlotWrapper` component for use in tests when we need to be able to pass content to `<slot/>`
- adds tests with high coverage for all new components (`Header`, `Nav`, and `NavItem`)
- adds the `@testing-library/user-events` dependency for more predictability when testing user interactions with components

## Screenshots

See [the Header story](https://641a2b7efd9419d13bbf8390-yugeewqjno.chromatic.com/?path=/story/example-header--default) on chromatic; the story should be responsive and show all three variations of the `NavItem`.

- [x] Fulfilled acceptance criteria on #1 (with support for mega menu as well)
- [x] Tests added to cover the change; run `npm run test:unit:coverage` for the coverage report (there were only a few branches I couldn't figure out how to cover in the `Header` and `NavItem`)

### Alternate solutions

As mentioned before, I would have liked for the `Nav` to be completely independent of the `Header`, but there didn't seem to be an easy way to do this given how the USWDS expects it to be structured.

In the future LDAF may want more control over the layout of the nav items in a mega menu; I could see them wanting to specify exactly which column and row an item appears in. If so we'll want to revisit this in the future and simplify how we layout mega menus, probably removing the `_.chunk` implementation.

Some of Courtney's mocks have included an image (a sort of featured item) in the mega menu in the first column; I opted to skip support for this for now, but I don't foresee this being too difficult to add in the future.

## Requested feedback

See self-review.

[LDAF-177]: https://ldaf.atlassian.net/browse/LDAF-177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/lib/components/Header/Header.test.ts
```diff
@@ -0,0 +1,21 @@
...
+describe("Header", () => {
+  it("renders", () => {
+    render(Header);
+    expect(screen.getByText("<Project title>")).toBeVisible();
+  });
+  it("opens nav on menu button click and closes it on close button click", async () => {
+    render(Header, { props: { navItems: [{ id: "1", link: "/", name: "test" }] } });
+    const navContainer = screen.getByLabelText("Primary navigation");
+    expect(navContainer).not.toHaveClass("is-visible");
+    await screen.getByText("Menu").click();
```
LouisFettet, line 16: I think since the `Header` component is using `on:click` handlers (as opposed to some of the `on:mousedown` and `on:keydown` in the `NavItem` component), we don't actually need to use `@testing-library/user-events` here to trigger the interaction.

src/lib/components/Header/Nav/NavItemType.ts
LouisFettet, line 1: This was the implementation I landed on which seemed to work best both in builds and tests. Happy to talk through alternatives here though.

src/lib/components/Header/Nav/NavItem.svelte
```diff
@@ -0,0 +1,108 @@
...
+      <span><slot /></span>
+    </button>
+    {#if megaMenu.length > 0}
+      <!-- Mega Menu Layout-->
+      <div
+        id="extended-mega-nav-section-{id}"
+        class="usa-nav__submenu usa-megamenu"
+        hidden={!isExpanded}
+      >
+        <div class="grid-row grid-gap-4">
```
LouisFettet, line 80: Wasn't sure if the `gap` here should be calculated based on the number of columns, so I just left this as the same in the example.

src/lib/components/Header/Nav/NavItem.test.ts
```diff
@@ -0,0 +1,109 @@
...
+
+import NavItem from "./NavItem.svelte";
+import TestSlotWrapper from "$lib/components/TestSlotWrapper.svelte";
+
+const generateMenuItems = (n: number) =>
+  [...Array(n)].map((_, i) => ({
+    id: `${i + 1}`,
+    name: `nav item ${i + 1}`,
+    link: `/${i + 1}`,
+  }));
```
LouisFettet, line 14: This is really similar to the `generateMenuItems` function in `Header.stories.ts`; I wasn't sure if it would make sense to move this into a shared util or not since it wouldn't be running on the actual site build (and I couldn't export it from the story since that clashes with how Storybook expects exports to work).

src/lib/components/TestSlotWrapper.svelte
```diff
@@ -4,6 +4,7 @@
   import type { ComponentType, SvelteComponentTyped } from "svelte";
 
   export let Component: ComponentType<SvelteComponentTyped>;
+  export let slotContent: string | Node;
```
LouisFettet, line 7: Unsure if `Node` is the right alternative here or not, but I figure we can update this as needed later.

src/stories/Header.stories.ts
```diff
@@ -1,26 +1,63 @@
...
+        id: genId(),
+        name: "MegaMenu 2",
+        megaMenuColumns: 3,
+        children: generateMenuItems(7, "MegaMenu 2"),
+      },
+      {
+        id: genId(),
+        name: "MegaMenu 3",
+        megaMenuColumns: 5,
+        children: generateMenuItems(22, "MegaMenu 3"),
```
LouisFettet, line 59: In the storybook UI, it seems like when we're generating a mega menu, any modifications to `children` will not update the component preview. This is probably an issue with the `NavItem` implementation, and while I don't think it's an immediate problem, this could cause an issue during hydration if we're fetching content from the CMS on page load. 

I tried a few quick fixes by using reactive declarations, but unfortunately the results were either no change or caused mega menus to render as basic menus, and I haven't looked much further into why yet. 

I figure this is something we can maybe look at later down the line (if so I'll probably add a comment about it `NavItem`), but if y'all think it's worth a closer look now, I can definitely continue digging.

Also, somewhat related, if y'all have any ideas on how to make it easier to modify the nav items in the story, let me know. I'm not a huge fan of how it's just appearing as a bunch of JSON right now.

LouisFettet (commented):
Self-review!

src/lib/components/Header/Nav/NavItem.svelte
LouisFettet: This is still relevant after the latest updates, but it's been moved: https://github.com/la-ldaf/ldaf-site/blob/8204289d1dc87b38f6895769a6a8f53e09382b55/src/lib/components/Header/Nav/NavMenu.svelte#L78

src/stories/Header.stories.ts
LouisFettet: Update: this was fixed by making `megaMenuColumns` reactive, which is something I had thought I had tried previously. Benaiah had done this in their review PR so doing something similar with 4abc75b got this working.

src/lib/util/chunk.ts
LouisFettet, line 1: @Benaiah I ended up reverting some of your changes here because I think there was some disconnect over what the second parameter was doing. Instead of `size` being the number of columns to divide the `input` into, it's actually the maximum length of a column. So by returning `[input]` when `size` was `1`, we were failing to divide things properly when we had a mega menu with N columns and items <= N.  For example, for a mega menu of three columns with three items, instead of getting `[['a'],['b'],['c']]`, it was returning `[['a','b','c']]`.

To make this a bit more clear I added `chunk.test.ts` as well as a comment here:
https://github.com/la-ldaf/ldaf-site/blob/f19d510ce7946586847cbad91042a6bee0b4ec98/src/lib/components/Header/Nav/NavMenu.svelte#L24-L25

src/lib/components/Header/Nav/NavMenu.svelte
```diff
@@ -0,0 +1,95 @@
...
+  // If we want more than one column, evenly divide the nav items into mega menu columns.
+  // TODO: We may want to let the CMS determine the position of each item in the future.
+  let megaMenuColumns: NavLinkType[][] = [];
+  $: {
+    if (columns > 1) {
+      // Calculate the maximum length of columns (maximum number of items in a column) and provide
+      //   to chunk as the size (second) parameter.
+      megaMenuColumns = chunk(children, Math.ceil(children.length / columns));
+    }
+  }
```
LouisFettet, line 28: @Benaiah I opted to change this from `childColumns` to `megaMenuColumns` to denote that it's only being used as a mega menu. I kept most of your changes around setting the default `columns` to `1` but checked that that value was greater than `1` before setting `megaMenuColumns`. I also edited and added comments here to make it clearer what was happening.

src/lib/util/chunk.ts
Benaiah: Ahh, ok! My bad on that. The important part of that change was the generic type, which you kept.

src/lib/components/Header/Nav/NavMenu.svelte
Benaiah: Sounds good, that's a good choice of variable name.

Benaiah (approved):
Looks great, thanks for all the work on this!

src/lib/components/Header/Header.scss
hinzed1127, line 1: This can get removed once the global import gets merged. If that happens first, I'll make a note to clean up other imports

src/lib/components/Header/Nav/Nav.svelte
```diff
@@ -0,0 +1,26 @@
+<script lang="ts">
+  import type { NavItemType } from "./types";
+
+  import NavItem from "./NavItem.svelte";
+
+  export let items: NavItemType[] = [];
+
+  // The index of the item that is currently expanded.
+  let expandedI: number | undefined = undefined;
```
hinzed1127, line 9: nit (i.e., change at your discretion): `expandedIndex` is more explicit and removes the need for a comment

hinzed1127 (approved):
Small things for follow-up plus a nit. Looks good!

src/lib/components/Header/Nav/Nav.svelte
LouisFettet: Agreed, I'll update this.

src/lib/components/Header/Header.scss
LouisFettet: Yeah I'll merge this as soon as tests pass on 9cbcded, so if you wouldn't mind updating that on your side I'd appreciate it!

PR #8, opened 2023-04-10
Set up automatic Dependabot version updates.

LouisFettet:
This adds a config file that will have Dependabot run on our `npm` packages weekly, and will open PRs that are automatically assigned to our engineering team. I set the commit prefix to `npm: ...` and it should automatically apply the `dependabot` and `dependencies` labels.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -17,8 +17,8 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
-      reviewers:
-        - "la-ldaf/ad-hoc-engineers"
+    reviewers:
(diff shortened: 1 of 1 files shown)


Benaiah (approved):
I'd love to be able to check if this works before merging but I'm not sure if we can - looks good to me based on the docs you linked tho.

I think we should add the `github-actions` ecosystem since we'll almost certainly be using it, you're welcome to include that here or we can do it in a followup when we add an action.

LouisFettet:
@Benaiah good call, I've added the GitHub actions ecosystem as well with a lot of the same config as npm. Merging and hoping this works as expected!

PR #9, opened 2023-04-11
LDAF-216: Refactor USWDS dependency to single global import

hinzed1127:
Jira ticket: [LDAF-216](https://ldaf.atlassian.net/browse/LDAF-216)

## Proposed changes

- Add sub-route that copies the [USWDS Documentation page template](https://designsystem.digital.gov/templates/documentation-page/#component-code)
- Use a single global import of USWDS
- Move shared components into top-level `+layout.svelte`

- [x] Fulfilled exit criteria 
- Refactoring of styles, tests not needed yet for this change

## Other details

[Gist summary of the different stylesheet sizes depending on import strategy   ](https://gist.github.com/hinzed1127/7baf4149a1a1e18d92e9b75fc3e9a907)

### Alternate solutions

- This _is_ the alternate solution. See full list of considered options in the [original approach taken](https://github.com/adhocteam/ldaf/pull/64).

### Possible drawbacks

- Individual page sizes might be marginally larger, since a single top-level global import of USWDS will cause PurgeCSS to remove all USWDS styles not used _overall_, not per individual page. This is likely a negligible different in the context of full bundle size, but we can revisit if page sizes go up disproportionately.

[LDAF-216]: https://ldaf.atlassian.net/browse/LDAF-216?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app.scss
+++ b/src/app.scss
@@ -1,5 +1,6 @@
 /* Write your global styles here, in SCSS syntax. Variables and mixins from the src/variables.scss file are available here without importing */
-@use "uswds-fonts";
-@use "uswds-global";
-@use "uswds-typography";
-@use "uswds-utilities";
+@use "uswds" with (
+  $theme-font-path: $theme-font-path,
+  $theme-image-path: $theme-image-path,
+  $theme-show-notifications: false
+);
--- a/src/routes/documentation/+page.svelte
+++ b/src/routes/documentation/+page.svelte
@@ -1,0 +1,80 @@
+<a class="usa-skipnav" href="#main-content">Skip to main content</a>
+
+<div class="usa-section">
+  <div class="grid-container">
+    <div class="grid-row grid-gap">
+      <div class="usa-layout-docs__sidenav desktop:grid-col-3">
+        <nav aria-label="Secondary navigation">
+          <ul class="usa-sidenav">
+            <li class="usa-sidenav__item"><a href="/">Parent link</a></li>
+            <li class="usa-sidenav__item">
+              <a href="/" class="usa-current">Current page</a>
+              <ul class="usa-sidenav__sublist">
+                <li class="usa-sidenav__item"><a href="/">Child link</a></li>
+                <li class="usa-sidenav__item">
+                  <a href="/" class="usa-current">Child link</a>
+                  <ul class="usa-sidenav__sublist">
+                    <li class="usa-sidenav__item">
+                      <a href="/">Grandchild link</a>
+                    </li>
+                    <li class="usa-sidenav__item">
+                      <a href="/">Grandchild link</a>
+                    </li>
+                    <li class="usa-sidenav__item">
+                      <a href="/" class="usa-current">Grandchild link</a>
+                    </li>
+                    <li class="usa-sidenav__item">
+                      <a href="/">Grandchild link</a>
+                    </li>
+                  </ul>
+                </li>
+                <li class="usa-sidenav__item"><a href="/">Child link</a></li>
+                <li class="usa-sidenav__item"><a href="/">Child link</a></li>
+                <li class="usa-sidenav__item"><a href="/">Child link</a></li>
+              </ul>
+            </li>
+            <li class="usa-sidenav__item"><a href="/">Parent link</a></li>
+          </ul>
+        </nav>
+      </div>
+      <main
+        class="
+            usa-layout-docs__main
+            desktop:grid-col-9
+            usa-prose usa-layout-docs
+          "
+        id="main-content"
+      >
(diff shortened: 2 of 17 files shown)


src/app.scss
```diff
@@ -1,5 +1,6 @@
 /* Write your global styles here, in SCSS syntax. Variables and mixins from the src/variables.scss file are available here without importing */
-@use "uswds-fonts";
-@use "uswds-global";
-@use "uswds-typography";
-@use "uswds-utilities";
+@use "uswds" with (
+  $theme-font-path: $theme-font-path,
+  $theme-image-path: $theme-image-path,
+  $theme-show-notifications: false
+);
```
hinzed1127, line 6: Now that this is just a global import, we don't need the workaround with vite injecting it via `css.preprocessorOptions.scss.additionalData`

src/routes/+layout.svelte
hinzed1127, line 1: I imagine this is similar to what we'll use the top-level layout when we switch to the actual design. It organizes things more sensibly with these demo pages in the meantime.

hinzed1127 (commented):
Self-review

src/routes/+layout.svelte
LouisFettet: Yeah this is very similar to what I did in my refactor of [the root layout](https://github.com/la-ldaf/ldaf-site/blob/8b860b446d3e574993ac7becc686d2a832024b4c/src/routes/%2Blayout.svelte) in my PR, so this LGTM! Only difference is that I think I included the `main` in my layout as well (you can either make this change here or we can handle it in mine).

LouisFettet (changes requested):
Most everything here looks good, but I'm seeing some issues with the `Link` component both in Storybook and when it's used on a page.  Seems like `src/lib/components/Link/Link.scss` is the main culprit, but I'm not sure if the declarations in `app.scss` will work in Storybook so we may need to add some sort of global decorator (or repurpose the `Prose` decorator).

Benaiah:
> I'm not sure if the declarations in `app.scss` will work in Storybook

Storybook has support for using a global CSS file, so assuming we can get it to work with SCSS this should be doable.

hinzed1127:
@LouisFettet [this](https://github.com/la-ldaf/ldaf-site/pull/9/commits/05fd310583c76e924bcc201c772574bbdcd8ff50) is all that was need to fix the Link component. I'm pretty sure the issue was due to `Link.scss` referencing variables/properties on the `uswds` module and storybook not being aware of it. Importing the global `app.scss` file in the storybook `preview.ts` file didn't fix anything, which probably has to do with how it compiles. I think we'll only have include that `@use` statement in storybook components specifically referencing `uswds`, so it shouldn't need to be done often.

LouisFettet:
@hinzed1127 sounds good. Couple of quick follow-up questions:
1. I assume the addition of the `@use` statement to `Link.scss` doesn't hurt the build size since we're still utilizing `PurgeCSS`, right?
2. I'm seeing the "USWDS Notifications" pop up in the console when running Storybook; any chance we could disable those? I _think_ we may be able to accomplish this with @Benaiah's suggestion to add `import "/src/app.scss";` to `/.storybook/preview.ts`

hinzed1127:
> 1. I assume the addition of the `@use` statement to `Link.scss` doesn't hurt the build size since we're still utilizing `PurgeCSS`, right?

Build size seems unaffected. I'm still seeing the single `_layout.*.css` file in the final build, size unchanged from before

> 2. I'm seeing the "USWDS Notifications" pop up in the console when running Storybook; any chance we could disable those? I _think_ we may be able to accomplish this with @Benaiah's suggestion to add `import "/src/app.scss";` to `/.storybook/preview.ts`

This one is strange. Importing `app.scss` in into the `preview.ts` doesn't get rid of the console notification, even though it's false in both `app.scss` and `Link.scss`

LouisFettet (approved):
LGTM! Might be good to get @Benaiah's approval as well since they've already spent a good bit of time looking into the related solutions.

Benaiah (approved):

LouisFettet (commented):
I think this is pretty much ready to go, just saw a redundant skipnav link.

src/routes/documentation/+page.svelte
```diff
@@ -0,0 +1,80 @@
+<a class="usa-skipnav" href="#main-content">Skip to main content</a>
```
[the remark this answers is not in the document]
hinzed1127: good catch!

PR #16, opened 2023-04-12
Review suggestions for #2

Benaiah:
- Makes `chunk` generic so that it doesn't infect variables with `any` types.
- Splits up `NavItemType` into `NavLinkType` and `NavMenuType`
- Splits up `NavItem` into `NavLink` and `NavMenu`
- Keeps "which, if any, menu is expanded" state in `Nav` component
- Uses events to signal expanding and closing the menu so that state can be kept in `Nav`

### DIFF
--- a/src/lib/components/Header/Nav/NavItem.svelte
+++ b/src/lib/components/Header/Nav/NavItem.svelte
@@ -1,108 +1,21 @@
 <script lang="ts">
   import type { NavItemProps, NavLinkType } from "./types";
 
-  import classNames from "$lib/util/classNames";
-  import chunk from "$lib/util/chunk";
-  import Link from "$lib/components/Link";
   import NavLink from "./NavLink.svelte";
+  import NavMenu from "./NavMenu.svelte";
 
   type $$Props = NavItemProps;
 
   export let id = "";
-  export let link = "";
-  export let isCurrent = false;
-  export let megaMenuColumns = 0;
+  export let link: string | undefined = undefined;
+  export let current = false;
+  export let columns = 0;
   export let children: NavLinkType[] = [];
-
-  let isParent = children.length > 0;
-  let isExpanded = false;
-
-  $: buttonClassNames = classNames(
-    "usa-accordion__button usa-nav__link",
-    isCurrent && "usa-current"
-  );
-
-  // Divide nav items into columns.
-  // TODO: We may want to let the CMS determine the position of each item in the future.
-  let megaMenu: NavLinkType[][] = [];
-  if (megaMenuColumns > 0) {
-    const columnLength = Math.ceil(children.length / megaMenuColumns);
-    megaMenu = chunk(children, columnLength);
-  }
-
-  const toggleSubMenu = (show?: boolean) => {
-    if (typeof show !== "undefined") isExpanded = show;
-    else isExpanded = !isExpanded;
-  };
-
-  // By default focusout on the container will trigger for all its children, but using 'self'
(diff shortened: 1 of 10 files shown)


src/lib/components/Header/Nav/NavItem.svelte
```diff
@@ -1,108 +1,21 @@
...
-  // We can't use on:click since it only triggers if the mousedown and mouseup events occur on the
-  //   same target, and on mobile the collapse of an accordion when it loses focus happens on
-  //   mousedown, sometimes shifting the elements so that mouseup misses the target.
-  // Instead, we'll use both mousedown and keydown to ensure everything functions accessibly on
-  //   both desktop and mobile screen sizes.
-  const handleDropdownKeyDown = (event: KeyboardEvent) => {
-    if (event.key === "Enter" || event.key === " ") toggleSubMenu();
-  };
+  export let link: string | undefined = undefined;
+  export let current = false;
```
LouisFettet, line 11: This needs to be changed in the story:https://github.com/la-ldaf/ldaf-site/blob/784dfc99e10cc383d6f4e39a03e1dbd7f8f6a589/src/stories/Header.stories.ts#L32

And the unit test:
https://github.com/la-ldaf/ldaf-site/blob/784dfc99e10cc383d6f4e39a03e1dbd7f8f6a589/src/lib/components/Header/Nav/NavItem.test.ts#L28

Honestly I probably should have checked that the element was getting the expected class name when it was set to `current`, so I'll probably handle that as a follow-up in my PR.

src/lib/components/Header/Nav/NavItem.svelte
```diff
@@ -1,108 +1,21 @@
...
+  export let columns = 0;
+  export let children: NavLinkType[] = [];
```
LouisFettet, line 13: This wasn't something I ever mentioned explicitly, but the reason for using the same type here was to support deep nesting in the menu in the future (even though my code in this PR didn't properly support it past one level).  This did make me remember to ask about whether we should be supporting this or not, so I posted [a question to our UX folks in Slack](https://adhoc.slack.com/archives/C04EL8QK5FU/p1681405708168349), and it seems like for simplicity's sake we should probably stick with just one level of nesting for now. 

So all that said, this works for now, but we may need to revisit it in the future.

src/lib/util/chunk.ts
```diff
@@ -1,9 +1,13 @@
 // https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_chunk
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-const chunk = (input: any[], size: number) => {
-  return input.reduce((arr, item, idx) => {
-    return idx % size === 0 ? [...arr, [item]] : [...arr.slice(0, -1), [...arr.slice(-1)[0], item]];
-  }, []);
+const chunk = <T>(input: T[], size: number): T[][] => {
+  if (size === 0) return [];
+  if (size === 1) return [input];
```
LouisFettet, line 5: Good additions here!

LouisFettet (approved):
Most everything here looks good. I had been thinking about splitting the menu and links into separate components but just wanted to get something out of the door, so that improvement is much appreciated. I also had wanted to give the event dispatcher a try but couldn't find a good use-case for it while everything was sitting in one file, so this should serve as a great example for future work.

There's just a couple of issues with updating `isCurrent` to `current` in the unit test and story, but I can handle that change in my PR if you want to go ahead and merge this as-is.

PR #17, opened 2023-04-12
LDAF-155: Add GitHub Action workflow for pull requests

hinzed1127:
Jira ticket: [LDAF-155](https://ldaf.atlassian.net/browse/LDAF-155)

## Proposed changes

- Adds a very barebones workflow to execute `npm run lint` and `npm run test:unit` on PR creation/updates

## Screenshots

None

- [ ] Workflow runs on each commit

## Requested feedback

[LDAF-155]: https://ldaf.atlassian.net/browse/LDAF-155?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,0 +1,28 @@
+name: "Linting and Testing"
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, ready_for_review]
+
+jobs:
+  lint-and-test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup node
+        uses: actions/setup-node@v3
+        with:
+          node-version: current
+          cache: npm
+
+      - name: Install
+        run: npm ci
+
+      - name: Link check
+        run: npm run lint
+
+      - name: Unit tests
+        run: npm run test:unit
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -2,7 +2,9 @@ name: "Linting and Testing"
 
 on:
   pull_request:
-    types: [opened, reopened, synchronize, ready_for_review]
+  #  by default, runs when a PR activity type
+  #  is `opened`, `synchronize`, or `reopened`
+  # (https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request)
 
 jobs:
   lint-and-test:
@@ -23,6 +25,9 @@ jobs:
 
       - name: Prettier & ESLint
         run: npm run lint
+
+      - name: svelte-check (diagnostic checks for unused CSS, Svelte A11y hints, and JS/TS errors)
+        run: npm run check
 
       - name: Unit tests
         run: npm run test:unit
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Install
         run: npm ci
 
-      - name: Link check
+      - name: Prettier & ESLint
         run: npm run lint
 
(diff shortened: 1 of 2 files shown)


.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,28 @@
+name: "Linting and Testing"
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, ready_for_review]
```
Benaiah, line 5: What's the motivation for including `opened` and `ready_for_review`?

Benaiah (changes requested):
Can we add `npm run check` as well?

.github/workflows/pull_request.yml
LouisFettet: I think by default it will run on `opened`, `synchronize`, and `reopen`. `synchronize` will be the main one that gets used most of the time but I don't think it triggers until a new commit is pushed to a PR, necessitating `opened` for new PRs. I could see how `reopened` and `ready_for_review` might not be entirely necessary, since I imagine the code shouldn't be changing since the last commit (and if it does need to change, that's when we merge `main` which should trigger `synchronize`). I'm new to configuring Actions, so let me know if I'm off-base at all here.

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,28 @@
...
+          cache: npm
+
+      - name: Install
+        run: npm ci
+
+      - name: Prettier & ESLint
+        run: npm run lint
+
+      - name: Unit tests
+        run: npm run test:unit
```
LouisFettet, line 28: Do we need to set `process.env.CI` manually anywhere in order to ensure it doesn't run in [watch mode](https://vitest.dev/guide/features.html#watch-mode)?

LouisFettet (commented):
It might also be useful to combine all the scripts that we want to run into just one npm script, e.g. `npm run test:ci`, although I can see how it would be useful to divide these checks into separate steps so that they're more clearly labeled.

.github/workflows/pull_request.yml
Benaiah: GitHub automatically sets it as of 2020: https://github.blog/changelog/2020-04-15-github-actions-sets-the-ci-environment-variable-to-true/

.github/workflows/pull_request.yml
Benaiah: Whoops! Yeah I meant to say "what's the motivation for including `reopened` and ready_for_review`, not `opened`. We can leave them in but I'm pretty sure `reopened` and `ready_for_review` are unnecessary since I'm pretty sure CI will run as normal on draft PRs. I'm also fine leaving them in and removing them if they cause trouble, up to @hinzed1127.

hinzed1127: Being new to GH actions, I based this off of the [vets-website PR workflow](https://github.com/department-of-veterans-affairs/vets-website/blob/4bd99dd65c152c1cfed545af9b71bed25484bdac/.github/workflows/pull-request.yml#L5). A closer look at the [docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) confirms Louis' thoughts (i.e., it runs on `opened`, `synchronize`, and `reopen` by default).  

I'm fine letting it go off by the defaults and just noting them as a comment in the file. `ready_for_review` would just get triggered after removing draft status from a PR, but it would still have checks run with every push, so the extra trigger seems redundant.

hinzed1127:
> I can see how it would be useful to divide these checks into separate steps so that they're more clearly labeled.

Yeah, that was my main rationale. Felt just a little cleaner

PR #18, opened 2023-04-13
Lazy Image component

Benaiah:
## Closes #12345 

Jira ticket: [LDAF-212](https://ldaf.atlassian.net/browse/LDAF-212)

## Proposed changes

- Adds an `Image` component that does the following:
  - If available, renders the mean color of the image as a background
  - If available, renders the [blurhash](https://blurha.sh/) of the image as a background (above the mean color)
  - By default, lazy loads images once they have been scrolled onto the page and fades them in once they have loaded. This process uses the following techniques, from most preferred to least:
    - `loading="lazy"`
    - `IntersectionObserver`
    - Regular non-lazy image loading
  - Allows for eager image loading using the same component with `loading="eager"`
- Uses the `Image` component on the main skeleton page:
  - Changes the hero image to use a preloaded `Image`
  - Adds a sample lazy `Image` to the text of the page

## Screenshots

<details>

Mean color background:

image

Blurhash background:

image

Fully loaded image:

image

Video of eager image loading (Chrome, hard refresh, throttled to "Fast 3G"):

https://user-images.githubusercontent.com/1425133/231640434-5866f577-e781-4cc7-be9e-508e8480e0cf.mov

</details>

- [x] Images continue to work as expected
- [x] Stories added
- [x] Tests added and passing

## Other details

This does not yet implement `srcset` and responsive images. Doing so at build-time would be prohibitively slow, so we plan to implement this using an image service such as Vercel's. Without that set up first it's difficult to develop and debug the responsive image part of this component, so we'll do that in a followup PR.

### Alternate solutions

We could do something simpler and remove the blurhash and mean-color backgrounds. This would also let us drop the dependency on Imgix, which we plan to use at build-time to generate the blurhashes and mean colors. (At runtime, regardless of which approach we use, we will use Vercel's image optimization for the best integration and edge deployment).

### Possible drawbacks

There should be no major drawbacks to this approach other than code size and potentially processor usage when generating blurhashes.

[LDAF-212]: https://ldaf.atlassian.net/browse/LDAF-212?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


LouisFettet:
Haven't gotten a chance to look at the code yet, but I'm getting an error when running both `npm run dev` and `npm run build` after a fresh `npm install`.

First issue seems like it may just be a warning, but I'm not sure:
```
@rollup/plugin-typescript TS1149: File name 'ldaf-site/src/lib/components/Image/drawBlurhash.ts' differs from already included file name 'ldaf-site/src/lib/components/Image/drawBlurHash.ts' only in casing.
  The file is in the program because:
    Root file specified for compilation
    Imported via "./drawBlurhash" from file 'ldaf-site/src/lib/components/Image/renderBlurhash.ts'
    Imported via "./drawBlurhash" from file 'ldaf-site/src/lib/components/Image/window.d.ts'
```

This is the error I get after the build fails:
```
[import-as-bundle-string-loader] Could not load ldaf-site/src/lib/components/Image/renderBlurhash.ts?bundlestring (imported by src/lib/components/Image/BlurhashRenderer.svelte): Unexpected token (Note that you need plugins to import files that are not JavaScript)
file: ldaf-site/src/lib/components/Image/drawBlurhash.ts:5:22
3: // The canvas should be sized to the expected rendering sized of the blurhash _before_ this function
4: // is called
5: export default (canvas: HTMLCanvasElement, blurhash: string) => {
                         ^
6:   const { width, height } = canvas;
7:   const pixels = decodeBlurHash(blurhash, width, height);
```

Benaiah:
Weird, I thought I had gotten those errors to stop happening! I'll try a clean build of it and see if I can figure it out.

src/lib/components/IntersectionObserver/key.ts
```diff
@@ -0,0 +1 @@
+export default {};
```
LouisFettet, line 1: Can you provide a bit of explanation somewhere about how this is getting used? I feel like I'm halfway to understanding how this is working with Svelte context, but a quick comment (or type declaration, if applicable) would help.

src/lib/components/Image/Image.svelte
```diff
@@ -0,0 +1,110 @@
...
+  let intersecting = false;
+
+  $: srcProps =
+    loading === "eager" && src
+      ? { src }
+      : !browser || !src
+      ? {}
+      : lazyImageLoadingSupport || !intersectionObserverSupport || intersecting
+      ? { src }
+      : {};
```
LouisFettet, line 42: nit: I think this might be more readable as an if / else block.

src/lib/components/Image/Image.test.ts
LouisFettet, line 1: Tests here look really good. I noticed that coverage on the `IntersectionObserver` seems rather low (specifically on `src/lib/components/IntersectionObserver/observe.ts`), which doesn't seem quite accurate to me based on the test cases here. Are there any additional test cases we could write here (or existing ones we could modify) to check for specific interactions with the intersection observer, e.g. that the image is `unobserved` after the image loads? Alternatively you could add a couple of tests specifically for the `IntersectionObserver` component. 

Feel free to disregard this suggestion if you feel like the tests here are already covering all our bases. As I've said before, coverage can sometimes be a useful metric, but it shouldn't hold up progress if we feel like we've tested a component thoroughly already.

LouisFettet (commented):
Everything seems to be working great when running locally both with `dev` and `build`. My comments are all pretty minor and focused mostly on adding code comments and improving readability. I've got one rabbit hole you could go down with testing, but from the test cases you've already written I'm not sure it's necessary.

src/stories/Image.stories.ts
```diff
@@ -0,0 +1,58 @@
...
+import Image from "$lib/components/Image";
+import BlurhashRendererDecorator from "./decorators/BlurhashRendererDecorator.svelte";
+import RootIntersectionObserverDecorator from "./decorators/RootIntersectionObserverDecorator.svelte";
+
+const srcOptions = { "sample image": sampleImage, none: undefined };
+const blurhashOptions = { "sample image blurhash": sampleImageBlurhash, none: undefined };
+const meanOptions = { "sample image mean color": sampleImageMean, none: undefined };
+
+const meta = {
+  title: "Image",
```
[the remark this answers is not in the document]
Benaiah: Agreed, I meant to make an issue about this now that the example stories have been removed but a quick-fix PR without an issue is fine.

Benaiah:
Had to squash this into one commit because the rebase was kind of a nightmare doing it one-by-one. The diff should be unaffected except for where I had to merge in changes to the layout or the main page.

src/lib/support.ts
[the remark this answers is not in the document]
Benaiah: I think `warn` could go in `util`, and I've moved it there, but I don't think `support` fits in there because it's a set of constants, not utility functions. We could split it up into multiple files in its own directory and/or rename it to `browserSupport` to make it clearer, but I don't think it fits in `util`.

Benaiah:
@LouisFettet ok, I went ahead and added a bunch of tests, both of `IntersectionObserver` itself and one of it actually setting the `src` on the `Image` component! I re-requested a review - let me know if there's anything else that you see that I should improve.

src/lib/components/IntersectionObserver/IntersectionObserver.test.ts
```diff
@@ -0,0 +1,83 @@
+import "@testing-library/jest-dom";
+import { render, screen, waitFor } from "@testing-library/svelte";
+import type { ComponentProps } from "svelte";
+import { vi, describe, it, expect } from "vitest";
+
+import IntersectionObserverMock, {
+  mock as intersectionObserverMock,
+} from "./__tests__/IntersectionObserverMock";
+const { intersect, observe, unobserve } = intersectionObserverMock;
```
LouisFettet, line 9: Got a minor linter warning here.

src/lib/components/IntersectionObserver/__tests__/IntersectionObserverTest.svelte
LouisFettet, line 1: Good call utilizing a `__tests__` dir for this. While you're here, would you mind making a similar change to `src/lib/components/TestSlotWrapper.svelte`, e.g. moving it to `src/lib/components/__tests__/TestSlotWrapper.svelte`? I know it's not quite related, but I figure it'd be easier than opening up a single file rename PR.

LouisFettet (approved):
Changes look great! You definitely went above and beyond what I was expecting for the testing here, so hopefully it wasn't too painful 😬 

I have a couple minor notes but otherwise I think this is good to go! Learned a lot from reviewing this PR; thanks for all your work here!

PR #22, opened 2023-04-14
Render Contentful rich text content

Benaiah:
Jira ticket: [LDAF-211](https://ldaf.atlassian.net/browse/LDAF-211)

## Proposed changes

- Adds a `ContentfulRichText` component that renders regular Contentful rich text.
- Adds a page at `/test-contenful-content` that uses the `ContentfulRichText` component.

## Screenshots

Screenshot is of [the Contentful rich text preview page](https://ldaf-git-benaiah-render-contentful-content-ldaf.vercel.app/test-contentful-content).

<details>

image

</details>

- [x] Contentful rich text example document renders correctly

[LDAF-211]: https://ldaf.atlassian.net/browse/LDAF-211?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/lib/__tests__/extendExpect.ts
```diff
@@ -0,0 +1,20 @@
+import { toMatchDOMNodes } from "./matchers";
+import { expect } from "vitest";
+
+interface CustomMatchers<R = unknown> {
+  toMatchDOMNodes(expected: HTMLElement | Node[] | string): R;
+}
+
+declare global {
+  /* eslint-disable-next-line @typescript-eslint/no-namespace */
+  namespace Vi {
```
Benaiah, line 10: This will need to be updated if we merge the vitest dependency update (#105) first.

src/routes/test-contentful-content/+layout.svelte
```diff
@@ -0,0 +1,3 @@
+<div class="grid-container">
```
LouisFettet, line 1: Talked about this in Slack, but adding the `usa-prose` class here fixes some of the styling to match USWDS (heading font, tables, etc.):
```suggestion
<div class="usa-prose grid-container">
```

package.json
```diff
@@ -76,6 +78,8 @@
   "type": "module",
   "dependencies": {
     "@uswds/uswds": "^3.4.1",
-    "fast-blurhash": "^1.1.2"
+    "@contentful/rich-text-types": "^16.0.3",
+    "fast-blurhash": "^1.1.2",
+    "lodash": "^4.17.21"
```
LouisFettet, line 83: I know lodash isn't a huge dependency, and I'm not entirely clear on if we're doing any tree-shaking automatically, but if we're not, we may want to find a way to just get the `lodash/escape` function that we're using here (and maybe do something with `lodash/chunk` while we're at it). 

I haven't looked into this all that much, but I think I've found two ways of going about this:
1. Install  [lodash.escape](https://www.npmjs.com/package/lodash.escape) and [lodash.chunk](https://www.npmjs.com/package/lodash.chunk). My guess is that there would be some overlap between these two but that might be the price of convenience.
2. Create a [custom build](https://lodash.com/custom-builds) using a CLI where we run `lodash include=chunk,escape` to generate a much smaller package. This would probably create the smallest dependency size, but might be a pain to maintain.

All that said, if you're seeing that we're only including relevant files for `lodash/escape` in the build, you can completely ignore everything I've said above 😅 . Might be nice to replace our `chunk` implementation with `lodash/chunk` while you're here though.

src/lib/components/ContentfulRichText/predicates.ts
```diff
@@ -0,0 +1,65 @@
...
+  isNode(n) && n.nodeType === BLOCKS.LIST_ITEM;
+
+export const isHyperlink = (n: unknown): n is Hyperlink =>
+  isNode(n) && n.nodeType === INLINES.HYPERLINK;
+
+export const isTable = (n: unknown): n is Table => isNode(n) && n.nodeType === BLOCKS.TABLE;
+export const isTableRow = (n: unknown): n is TableRow =>
+  isNode(n) && n.nodeType === BLOCKS.TABLE_ROW;
+export const isTableHeaderCell = (n: unknown): n is TableHeaderCell =>
+  isNode(n) && n.nodeType === BLOCKS.TABLE_HEADER_CELL;
```
LouisFettet, line 63: Any idea why this is coming up as uncovered in the coverage report? The other uncovered lines seem like other common issues we've seen with the reporter (exporting and assigning values mostly), but we definitely have table headers in the test content; is markdown not rendering them to HTML appropriately?

This is super unimportant, especially since I think you're already doing a great job of covering all our bases here; don't sweat spending more than a couple minutes looking at this!

src/lib/components/ContentfulRichText/__tests__/documents/markdown-document.md
```diff
@@ -0,0 +1,120 @@
...
+1. Lorem ipsum dolor sit amet
+2. Consectetur adipiscing elit
+3. Integer molestie lorem at massa
+
+4. You can use sequential numbers...
+5. ...or keep all the numbers as `1.`
+
+Start numbering with offset:
+
+57. foo
```
LouisFettet, line 71: Dumb question, but we're not expecting this to start the count at 57, right?

src/lib/components/ContentfulRichText/__tests__/documents/.generate-documents-from-markdown.cjs
```diff
@@ -0,0 +1,82 @@
+// Run this file to generate JSON rich text files from every markdown file in this directory
```
LouisFettet, line 1: Just curious, is this something we should be running as part of test setup instead of checking the `.md.html` and `.md.json` files into source?

LouisFettet (commented):
Got a few comments / questions (all of which are probably not blockers), but this looks good! Excited to plug in our CMS content and see the magic happen.

package.json
Benaiah: `lodash` exports a separate entrypoint for every function. If we imported `escape` like `import { escape } from "lodash"` we would have to verify that tree-shaking was working properly (it usually doesn't), but by importing it like `import escape from "lodash/escape"` we are ensuring that we are importing the minimal bundle to support `escape` (in fact, this is what plugins that claim to improve your lodash tree-shaking do, they rewrite imports to the `import defaultExport from lodash/*` form). See [this page in the Lodash docs](https://lodash.com/per-method-packages) for more details.

Benaiah: I've replaced our custom "chunk" function with `lodash/chunk` as you suggested.

src/lib/components/ContentfulRichText/predicates.ts
Benaiah: It's because `@contentful/rich-text-from-markdown`, which is where all our test cases come from, is buggy and only renders table rows and cells, not table header rows and cells, even though the latter are defined in `@contentful/rich-text-types`. You can confirm this by running `node src/lib/components/ContentfulRichText/__tests__/documents/.generate-documents-from-markdown.cjs` yourself and confirming that the generated documents have no table header cells in JSON nor HTML.

Benaiah: See https://github.com/contentful/rich-text/issues/352 for this issue.

src/lib/components/ContentfulRichText/__tests__/documents/.generate-documents-from-markdown.cjs
Benaiah: good point, I've made this part of the `pretest` script.

src/lib/components/ContentfulRichText/__tests__/documents/markdown-document.md
Benaiah: It's not supported by `@contentful/rich-text-from-markdown`, so I don't see a need for us to support it unless we find out we have to.

src/lib/components/ContentfulRichText/predicates.ts
LouisFettet: Good to know it's not an issue on our side! I would usually recommend leaving a comment to remind us why we're seeing this, but I would assume that this will eventually be fixed with a dependency bump so I'm cool with leaving everything as-is. Thanks for looking into this!

src/routes/test-contentful-content/+layout.svelte
```diff
@@ -0,0 +1,3 @@
+<div class="usa-prose grid-container">
```
LouisFettet, line 1: After looking at this a bit more I think we may also want to include the `usa-section` class to get the spacing right:
```suggestion
<div class="usa-section usa-prose grid-container">
```

LouisFettet (approved):
Latest changes all look good! I have one more minor suggestion but I think this is good to go!

PR #50, opened 2023-04-19
Add Site Header Logo/Title Component, overall styles

hinzed1127:
Jira tickets: [LDAF-225](https://ldaf.atlassian.net/browse/LDAF-225), [LDAF-226](https://ldaf.atlassian.net/browse/LDAF-226), [LDAF-227](https://ldaf.atlassian.net/browse/LDAF-227)

## Proposed changes

- Adds header with LDAF logo+title
- Adds flat and "shimmering" version of the header, TBD which we use (assets can be updated later)
- Adds beets photo background asset
- Updated styles 
- Encapsulates most of the visual changes within a `<Title />` component (save for the secondary nav bits that remain in `<Header />`)
- Adds the currently proposed [color palette](https://www.figma.com/file/oGKbyCnCRRdNzLYbiags93/LDAF-Component-Library-USWDS-3.0.2?node-id=2178-2218&t=kq2NK1EyeLmdQ38b-0) to `variables.scss` to have a globally accessible set of color variables
- adds in scaffolding for populating content from Contentful in title and secondary nav components
## Screenshots
**Narrow/Mobile View: Menu Closed**
narrow-view-menu-closed

**Narrow/Mobile View: Menu Open**
narrow-view-menu-open

**Desktop View**
wide-view

- [x] Link to homepage on logo (or title, when logo is hidden) click
- [x] images load normally (no blurhash)
- [x] matches proposed designs
- Not really any functional changes, purely visual, so no tests added yet. Updated a failing <Header /> test to use ARIA roles

## Requested feedback

- I tried to add a minimum of class selectors to be able to style things. I initially settled on prefixing the majority of selectors with `ldaf-`. This makes it clear where we're using our styles vs USWDS, as well as making things more easily searchable. I'm open to other ideas on the best way to achieve those goals

[LDAF-226]: https://ldaf.atlassian.net/browse/LDAF-226?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[LDAF-225]: https://ldaf.atlassian.net/browse/LDAF-225?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LDAF-227]: https://ldaf.atlassian.net/browse/LDAF-227?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/lib/components/Header/Header.test.ts
```diff
@@ -7,7 +7,8 @@ import Header from "./Header.svelte";
 describe("Header", () => {
   it("renders", () => {
     render(Header);
-    expect(screen.getByText("<Project title>")).toBeVisible();
+    // Implicit ARIA role for the HTML <header> element
+    expect(screen.getByRole("banner")).toBeVisible();
```
hinzed1127, line 11: This test failed with my changes, so this is both a fix and a change to something that should be more future-proof. I also tried to follow the [recommended precedence for queries by Testing Library](https://testing-library.com/docs/queries/about/#priority)

src/lib/components/Header/Title/Title.svelte
hinzed1127, line 1: this probably needs either:
1. its own `+Title.server.ts` file to handle text values, OR
2. props to be passed, those coming from `+layout.server.ts`
are there assumptions we can currently make about the shape of the data getting passed in for that?

src/variables.scss
```diff
@@ -1,3 +1,71 @@
 /* Variables and mixins declared here will be available in all other SCSS files */
 $theme-font-path: "@uswds/uswds/fonts";
 $theme-image-path: "@uswds/uswds/img";
+
+/* Design System Color Palette */
```
hinzed1127, line 5: [Figma Reference](https://www.figma.com/file/oGKbyCnCRRdNzLYbiags93/LDAF-Component-Library-USWDS-3.0.2?node-id=2178-2218&t=kq2NK1EyeLmdQ38b-0)

src/lib/components/Header/Header.svelte
```diff
@@ -38,10 +35,18 @@
         <Icon name="close" alt="Close" size={3} />
       </button>
 
+      <a class="ldaf-logo-link" href="/" title="<Project title>">
+        <img class="ldaf-logo-compact" src={ldafLogo} alt="LDAF Logo" />
+      </a>
+      <div class="ldaf-agency-commissioner">
+        <span>Mike Strain, DVM</span>
+        <span class="text-italic">Commissioner</span>
+      </div>
```
hinzed1127, line 44: Same question/sentiment here as in `Title.svelte`

hinzed1127 (commented):
self-review

src/lib/components/Header/Title/Title.svelte
LouisFettet: I think it's going to be pretty difficult to determine how we want to load in content until Contentful is set up, but I would hope that we would be able to make requests per-component (as opposed to one big series of requests in `+layout.server.ts`) without impacting build speed. I'd go with the `+Title.server.ts` approach here and structure the data however you'd like, since it will need to change regardless in the future.

Benaiah: We're not going to be able to do per-component fetching of data in any place we want SSR; all data for SSR rendering must be fetched in a `+page.server.ts` or `+layout.server.ts`. That said, we can separate out the data-fetching functions by making something like `Title.server.ts` that can live in this folder which exports a `loadTitle` function that loads the data necessary to render this component, then call it in the `+layout.server.ts` of the layout that uses this component.

Benaiah (changes requested):
Regarding the CSS additions, can we do one of the following:

1. Match the [BEM](https://getbem.com/introduction/)-style classes of USWDS when making new `ldaf-` classes? e.g., we have `.ldaf-secondary-nav.usa-nav__secondary` in this PR when it seems like it would be cleaner to use `.ldaf-nav__secondary.usa-nav__secondary`. This will make these classes much less likely to collide with other global classes in the future.

2. If you'd rather not come up with a bunch of BEM classes, instead move these styles into a scoped `<style lang="scss">` block in `src/lib/components/Header/Header.svelte` and then use whatever classes you want (as long as they aren't existing global classes), since they'll be scoped to that specific component.

src/lib/components/Header/Title/Title.svelte
LouisFettet: Oh that seems like a pretty good solution then. In that case we could do something similar for the nav items, i.e. create a `NavItem.server.ts` that `+layout.server.ts` uses so that the layout file doesn't become a behemoth (although don't worry about handling that in this PR).

Benaiah (approved):
LGTM! Thanks for all the work on this!

src/lib/components/Header/Header.scss
```diff
@@ -0,0 +1,67 @@
...
+}
+
+.ldaf-header .ldaf-nav__secondary {
+  .usa-input {
+    border-radius: 5px 0 0 5px;
+  }
+
+  a {
+    color: $primary-light;
+  }
```
LouisFettet, line 13: I think this will override the `:visited` state; I think this makes sense for this case, but just wanted to make sure it was intentional.

src/lib/components/Header/Header.svelte
```diff
@@ -39,17 +39,25 @@
...
-          <li class="usa-nav__secondary-item">
-            <a href="/news">Grants and funding</a>
-          </li>
-          <li class="usa-nav__secondary-item">
-            <a href="/documentation" class="usa-current">Licensing and permits</a>
-          </li>
+          {#each secondaryNavItems as item, i (item.id)}
+            {@const { name, link } = item}
+            <li class="usa-nav__secondary-item">
+              <a href={link}>{name}</a>
```
LouisFettet, line 58: We should use `Link` here instead, mainly because in the future I expect we'll probably be adding some sort of analytics events to that component that wouldn't get captured if we just used an `<a>`.

You could also set `alternate={true}` to get the right color by default on desktop, but this might not be the right call since we're styling mobile-first and would still need to override the color on mobile.

src/lib/components/Header/Title/Title.svelte
```diff
@@ -0,0 +1,32 @@
+<script lang="ts">
+  import "./Title.scss";
+  import ldafLogo from "$lib/assets/ldaf-flat-logo-transparent.png";
+  import type { SiteTitleType } from "./types";
+  export let siteTitle: SiteTitleType;
+</script>
+
+<div class="usa-logo ldaf-logo__wide" id="basic-logo">
+  <a href="/" title="<Project title>">
```
LouisFettet, line 9: We should either add a `TODO` here so we don't forget to replace this, or create a `title` that we can throw into `Title.server.ts`. Based on [this a11y guidance](https://www.w3.org/WAI/tutorials/images/functional/#example-1-image-used-alone-as-a-linked-logo) and since the rest of the descriptive text is not part of the link, I'm thinking we should set the `title` here to something like `"Louisiana Department of Agriculture and Forestry Home"`.

LouisFettet (commented):
This looks really great! Just have a handful of relatively minor comments. Only thing which might take a lift is my comment about the media queries, and I think it's fine if we handle that in a follow-up if there's not a quick solution there.

src/lib/components/Header/Header.svelte
```diff
@@ -19,15 +25,9 @@
...
-<header class="usa-header usa-header--extended">
-  <div class="usa-navbar">
-    <div class="usa-logo" id="basic-logo">
-      <em class="usa-logo__text">
-        <!--TODO: Replace with content from CMS. -->
-        <a href="/" title="<Project title>"> &lt;Project title&gt; </a>
-      </em>
-    </div>
-
+<header class="ldaf-header usa-header usa-header--extended">
```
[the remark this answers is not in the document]
hinzed1127: Had to do a quick confidence check on this, myself, but it doesn't affect appearance, since[ the actual rulesets themselves determine precedence](https://css-tricks.com/the-order-of-css-classes-in-html-doesnt-matter/), not class declarations.

I don't have strong feelings on keeping to any particular best practice here, since we're hopefully staying minimal enough on utility classes that the number of any classes on any single element won't get too large

src/lib/components/Header/Header.scss
```diff
@@ -0,0 +1,67 @@
+.ldaf-header {
+  background: linear-gradient(to right, transparent, #00284d 30%),
```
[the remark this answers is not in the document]
hinzed1127: good catch, I thought I caught all of the colors in variables, but totally missed this one

src/lib/components/Header/Header.scss
hinzed1127: Yeah the secondary nav links are colored differently, I think to account for the [background]([https://](https://www.figma.com/file/oGKbyCnCRRdNzLYbiags93/LDAF-Component-Library-USWDS-3.0.2?node-id=2196-11223&t=1yfrsbwflqSDvces-0)). Idk if we want to have a `:visited` state for those, but something to make @getpunched aware of if in case we do

src/lib/components/Header/Header.scss
```diff
@@ -0,0 +1,67 @@
...
+    }
+  }
+
+  .ldaf-commissioner__compact,
+  .ldaf-logo__compact {
+    display: none;
+  }
+}
+
+@media all and (max-width: 63.99em) {
```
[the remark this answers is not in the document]
hinzed1127: Good thought. I arrived at those units via inspected styles getting applied via USWDS anyway, so makes sense to try and utilize whatever they've got built in already whenever possible. This hopefully shouldn't be a difficult update, so looking into it...

src/lib/components/Header/Title/Title.svelte
hinzed1127: Based on that [first example](https://www.w3.org/WAI/tutorials/images/functional/#example-1-image-used-alone-as-a-linked-logo), does it make sense to just forgo the `title` entirely and just update the `alt` attribute on the `img`? I'm out of my element here, so I don't know what the best practice is.

src/lib/components/Header/Header.svelte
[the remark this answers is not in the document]
LouisFettet: Wow, what a trip-up on my part. It's called cascading for a reason 🤦

All good here, disregard!

src/lib/components/Header/Header.scss
LouisFettet: Sounds good to me!

src/lib/components/Header/Title/Title.svelte
LouisFettet: Ah yep, good call. +1 to remove the `title` from the `<a>` and add an `alt` to the ``.

LouisFettet (approved):
Looks great! Just need to make that adjustment with the logo alt text and you should be good to :shipit:

src/lib/components/Header/Title/Title.svelte
hinzed1127: to your point, though, the [compact title below this](https://github.com/la-ldaf/ldaf-site/pull/50/files#diff-0f06d8a7ccb1b7bf0eb8bb2cdce882d3e36e483c6fbc8b32b3d641f46a0ae9a9R26) could probably still have a title value served from the CMS

LouisFettet:
After merging the CI changes, you'll need to merge `main` in order to get the checks passing.

PR #54, opened 2023-04-21
Icon: lots of improvements

Benaiah:
## Proposed changes

- Inline icons
- Don't base64 inline SVG data URLs
- `Icon` component takes `src` instead of `name`
- Icon URLs and names can now be imported from `$icons/NAME` using another new Vite plugin
- All icons are now available instead of just those in `usa-icons`
- Icons are lazily imported in Storybook
- Build time impact of `Icon` decreased
- Bundle size of main layout JS file massively decreased
- Number of requests required for homepage decreased due to inlining

## Screenshots
 
N/A, everything looks the same as before.

- [x] Icons continue to render correctly on skeleton site
- [x] Icons can still be interacted with in Storybook

### Alternate solutions

1. We could continue using our current solution of eagerly loading every icon in `usa-icons` into the `Icon` component as dataurls using the existing SVG plugin. This has a number of downsides:
  - This imports icon dataurls as base64 dataurls, which are larger and don't compress as well as escaped-SVG-source dataurls.
  - This only allows access to a subset of the SVG icons available in USWDS and used on the skeleton site. Other icons have to be handled without the Icon component.
  - This impacts both build speed and bundle size negatively and significantly.

2. We could add a component generator to `vite-plugin-ldaf-icons` and import specific components
  - I tried this, but I couldn't figure out a way to avoid duplicating the prop types between the `Icon` component and the `declare module` that declared the types of the generated components
  - This has a lot more impact on build speed than the solution I ended up going with in this PR
  - Using `svelte.compile` this way is error-prone since it's a different Svelte compiler entry point than the rest of the application

## Requested feedback

Let me know wherever it's unclear on how this is working so that I can add comments explaining it. It's hard to know where to add those as the person who wrote the PR in the first place.


src/lib/components/Icon/Icon.svelte
```diff
@@ -1,31 +1,27 @@
...
   // If you want to specify an icon size, use one of the component’s size variants."
   export let size: 3 | 4 | 5 | 6 | 7 | 8 | 9 | undefined = undefined;
-  export let className = "";
 
-  // As a default value, make the alt just a more titleized version of the file name,
-  // e.g., "arrow_upward" becomes "Arrow Upward"
-  export let alt = name
-    .split("_")
-    .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
-    .join(" ");
```
LouisFettet, line 19: Is there another way we can provide a default here? Could we do something similar to what we had here by extracting the icon name out of the `src` string?

src/lib/components/Icon/Icon.svelte
```diff
@@ -1,31 +1,27 @@
...
+  export { className as class };
+
+  export let alt = "";
+  export let src: string;
+  export let plain = false;
 
-  if (!icons[name]) {
-    throw new Error(`${name} is not a valid icon!`);
-  }
+  if (plain && size) console.warn('"size" prop has no effect when "plain" is set');
```
LouisFettet, line 19: Might be mistaken, but isn't `plain = true` the same as `size = undefined` currently? If so I feel like it will be easier to omit `size` from props rather than remember to add `plain`.

src/lib/components/Icon/Icon.test.ts
```diff
@@ -1,24 +1,20 @@
...
     expect(screen.getByRole("img")).toHaveClass("usa-icon");
   });
   it("renders with a different size", () => {
-    render(Icon, { props: { name: "rss_feed", size: 6 } });
-    expect(screen.getByAltText("Rss Feed")).toHaveClass("usa-icon");
+    render(Icon, { props: { src: rssIcon, alt: "RSS Feed", size: 6 } });
+    expect(screen.getByAltText("RSS Feed")).toHaveClass("usa-icon");
     expect(screen.getByRole("img")).toHaveClass("usa-icon");
-    expect(screen.getByAltText("Rss Feed")).toHaveClass("usa-icon--size-6");
+    expect(screen.getByAltText("RSS Feed")).toHaveClass("usa-icon--size-6");
```
LouisFettet, line 18: Nit: since nothing is changing, we should only need to run the `getByAltText` query once and can reuse the result (same was true for the original test).

LouisFettet (commented):
Besides below minor comments, I'm seeing a couple of issues with storybook:
1. With `npm run storybook`, the icon sizes aren't displaying properly.
2. After `npm run build-storybook`, icons are not rendering at all (possibly because the plugin isn't running as part of that build?).

src/lib/components/Icon/Icon.svelte
Benaiah: `plain` is not the same as `size={undefined}`. `plain` is used to omit _all_ the default classes including `usa-icon`, whereas `size={undefined}` only omits `usa-icon--size-SIZE`.

src/lib/components/Icon/Icon.svelte
Benaiah: I don't think providing a default `alt` is a good idea to begin with. The names of the icons vary a lot, so making sure that all of the autogenerated labels are meaningful is already going to be difficult, and the choice of alt text for an icon is very situational. For instance, if there is no accompanying text, the icon should have alt text describing what the icon does, not what it visually looks like. Another example would be something like the search icon which appears right next to a "Search" label and so should have `alt=""`.

Benaiah:
>     1. With `npm run storybook`, the icon sizes aren't displaying properly.

This is an easy fix - we need to `import "../src/app.scss" in ".storybook/preview.ts`.

>     2. After `npm run build-storybook`, icons are not rendering at all (possibly because the plugin isn't running as part of that build?).

This is much less easy. I didn't think of this when setting up the icon stories. I might have to do an eager import of all the icons in the Storybook build to get this to work when it's built statically.

Benaiah:
@LouisFettet storybook should be working now in both dev and prod using `import.meta.glob`! Let me know if there's any more changes I need to make.

src/lib/components/Icon/Icon.svelte
LouisFettet: Ah okay, I missed that `plain` would omit `usa-icon` when reading. That makes more sense then, since there are cases where we want the SVG icon but don't want USWDS to apply the icon styling.

src/lib/components/Icon/Icon.svelte
LouisFettet: Right, good point. For some reason I thought we had added the default `alt` due to guidance from the USWDS but that doesn't seem to be the case. You're right that we don't always want an `alt` attribute on these images. Disregard then!

LouisFettet (approved):
Latest changes all look good to me! I think it would be good to get a quick review from @hinzed1127 as well since he worked on the icons originally.

Only thing that needs to get fixed before merge are some typescript errors in CI. As long as these are minor I don't think this needs a re-review.

src/lib/components/Link/Link.scss
```diff
@@ -1,4 +1,4 @@
-@use "uswds" with (
+@use "uswds-core" as uswds with (
```
hinzed1127, line 1: good catch, hopefully thats speeds things up a bit locally

vite-plugin-ldaf-icon/index.ts
hinzed1127, line 1: this is a really slick addition

hinzed1127 (approved):
looks good to me!

PR #57, opened 2023-04-21
Image: support for responsive images

Benaiah:
## Proposed changes

- Renders the `Image` component using a `picture` element with `source` children wrapping the `img` tag that was already present.
- Uses `vite-imagetools` to generate mobile and webp versions of the two sample images on the skeleton site

In production, we will use the Imgix APIs at build time to generate blurhashes and mean colors, and the Vercel and/or Contentful image APIs at runtime to generate different sizes and formats for visitors to the site. In the meantime, we want our components to support this in the future, so we use `vite-imagetools` to generate some samples of mobile-sizes and webp images to evaluate site performance with.

<details>

<summary>Mobile Lighthouse report after this PR</summary>

image

</details>

- [x] Images continue to render correctly
- [ ] Tests added to cover the change

## Requested feedback


LouisFettet (commented):
LGTM on initial review; only suggestion for now would be to try to store our magic number of `412` for the `width` somewhere. AFAIK we can't do something like:
```js
import heroImgMobile from `@uswds/uswds/img/hero.jpg?quality=${QUALITY}&w=${WIDTH}&imagetools`;
```
But we could at least use a constant when setting the `sources` prop on `Image`.

All that said, if `412` is really just a placeholder, I'd be fine with just leaving a few `TODO` comments here and there to note that we need to revisit the value. Let me know what you think!

hinzed1127 (approved):
I like @LouisFettet's idea of defining some size constants, but otherwise looks good!

Benaiah:
> LGTM on initial review; only suggestion for now would be to try to store our magic number of `412` for the `width` somewhere. [..] But we could at least use a constant when setting the `sources` prop on `Image`.
> 
> All that said, if `412` is really just a placeholder, I'd be fine with just leaving a few `TODO` comments here and there to note that we need to revisit the value. Let me know what you think!

412 is very much a placeholder (and is based on nothing except "whatever the lighthouse width was" so that I could see what the best possible Lighthouse score would be after these changes). We'll handle actually getting the resized images from Vercel/Imgix in followup PRs. Good call to leave some TODO comments indicating that we should do that.

I've added a data file at `$lib/constants/images.json` that contains two lists of `sizes` and `formats` to include when generating source sets. I've also added a function that generates a sources object from a fixed list of images for testing with sample images.

These changes are significant enough that I'm going to re-request review.

LouisFettet (commented):
Two questions:
1. I'm seeing a rather low mobile performance rating on Lighthouse when running `npm run build && npm run preview`; are you doing something different to get the results you're seeing or did something change?
2. Did you want to use that `412` constant in `src/lib/components/landingPage/Hero.svelte` and `src/routes/+page.svelte`? Or did you intentionally only use it for generating sources?

LouisFettet (approved):
> The mobile lighthouse performance score in the screenshot was obtained with the following steps:
> 
> * Open https://ldaf-git-benaiah-responsive-images-ldaf.vercel.app/ in Chrome
> * Disable the Vercel preview comments and refresh the page
> * Open the developer tools, switch to the Lighthouse tab, and run the mobile "navigation" report.
> 
> This process consistently gets me a lighthouse score between 95 and 99 on my machine.

That's odd, performing the same steps I'm getting low 70s in performance due to "First Contentful Paint" and "Largest Contentful Paint" being in the red. It's entirely possible that one of my extensions is mucking things up though.

Regardless, this is still an improvement and should be good to go as soon as conflicts are resolved.

PR #59, opened 2023-04-21
LDAF-237 Improve end-to-end testing and run the tests in CI

### LINKED ISSUE #58: Add end-to-end testing to CI
Jira ticket: [LDAF-237](https://ldaf.atlassian.net/browse/LDAF-237)

## Problem statement

There are some test cases with SvelteKit that can't be replicated with our component unit tests. For these cases, we'll want to run end-to-end tests to ensure that everything is working as expected. These tests should run on every commit as well as on-merge to `main`.

## Acceptance criteria

- [x] End-to-end tests run on every commit in a PR.
- [x] End-to-end tests run on every commit to `main`.
- [ ] End-to-end tests do not block merging.

As part of this work we may also want to revisit some of our testing scripts (e.g. use the new `--ui` flag for our end-to-end tests for local development).

[LDAF-237]: https://ldaf.atlassian.net/browse/LDAF-237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

LouisFettet:
## Closes #58

Jira ticket: [LDAF-237](https://ldaf.atlassian.net/browse/LDAF-237)

## Proposed changes

- Renames existing `pull_request` workflow to `lint_and_unit_test` ("Linting and Unit Testing").
- Creates new `e2e_test` workflow that will run on `pull_request` and on `push` to `main`.
  - I figured it would make sense to break these out by the job they're doing rather than their trigger, that way we're not duplicating config between the workflows for pull requests and pushes to `main`. It also makes it easier to specify that this workflow won't block merging.
- Adds additional configuration settings to our end-to-end tests:
  - Adds UI mode as an option for local development with`npm run test:e2e:ui` (includes creating results you can view after a run).
  -  Don't rebuild the site if you already have a preview running.
  - Sets some defaults when running in CI (retry once, run with two workers, always build from scratch).
  - Adds HTML reports (in addition to existing CLI reporting).
- Update our ignore files to ignore test results and reports.

- [x] End-to-end tests run on every commit in a PR.
- [x] End-to-end tests run on every commit to `main`. (Will need verification after merge.)
- [ ] End-to-end tests do not block merging. (Will need to edit this in project settings after merge.)

## Requested feedback

Does the new directory structure, with `e2e-tests` at the top level and `report` and `results` as subdirectories, make sense? Happy to rework this!

[LDAF-237]: https://ldaf.atlassian.net/browse/LDAF-237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/e2e_test.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Build site and run end-to-end tests
+        run: npm run test:e2e
+
+      - name: "Upload test report"
+        uses: actions/upload-artifact@v3
+        if: always()
+        with:
+          name: e2e-test-report
+          path: e2e-tests/report/
+          retention-days: 30
```
LouisFettet, line 39: Didn't know about this option until I was working on this; figure we'll probably do the same for unit test coverage reports. Let me know if y'all have any feelings about the `retention-days` limit I've set; a month feels sufficient but I'm down to increase it.

It's a bit of a pain because you have to download the file and then run `npx playwright show-report` in order to view it (instead of GitHub hosting it and just clicking a link), but since this shouldn't be a problem too often I figure this is fine for now.

.github/workflows/lint_and_unit_test.yml
```diff
@@ -17,7 +16,7 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v3
         with:
-          node-version: current
+          node-version-file: ".nvmrc"
```
LouisFettet, line 19: This works pretty nicely and is one less thing we'll need to update the next time we upgrade our Node version.

From the workflow logs:
```
Node version file is not JSON file
Resolved .nvmrc as lts/hydrogen
Attempt to resolve LTS alias from manifest...
Found in cache @ /opt/hostedtoolcache/node/18.16.0/x64
```

package.json
```diff
@@ -7,8 +7,9 @@
     "dev": "vite dev",
     "build": "node scripts/pre-build.js && vite build",
     "preview": "vite preview",
-    "test": "npm run lint && npm run test:unit",
+    "test": "npm run lint && npm run check && npm run test:unit:coverage",
```
LouisFettet, line 10: Unrelated, but figured we should include `npm run check` in our default `test` command.

package.json
```diff
@@ -7,8 +7,9 @@
...
     "test:e2e": "playwright test",
+    "test:e2e:ui": "npm run test:e2e -- --ui",
```
LouisFettet, line 12: It's easy enough to just add the `-- --ui` flag but it's a couple less keystrokes, and I feel like I'll forget about this option otherwise 😅

LouisFettet (commented):
Self-review!

As mentioned in the PR description, I'd also like feedback on the `e2e-test` directory placement.

.github/workflows/e2e_test.yml
Benaiah: This looks good for now! I think it would be nice as a followup to upload it to the final site preview under something like `/.meta/playwright-report` so we can link to the report directly within the PR and keep the report around for as long as the deployment exists. We should be able to use the artifact upload/download to do so, so this PR sets us up to do that in the future.

LouisFettet: Yeah I'll be handling adding code coverage reports to the build in #67 / #68, so I might handle this there as well.

I'm curious how you think we should go about doing this; brute-force would be to run the unit tests and end-to-end tests as part of the Vercel build command (that way all the necessary files are immediately available to Vercel), but there's got to be a better way right? Ideally we'd be able to fetch artifacts from GitHub as part of the Vercel build process but I suppose we'd have to delay the Vercel build, maybe only triggering it once tests are completed. Maybe there's a simpler solution here that I'm just not seeing?

package.json
Benaiah: I'm ok with adding `npm run check`, but I don't think we should make this run `test:unit:coverage` instead of `test:unit` because `test:unit:coverage` disables the default behavior of watching the tests and source and rerunning tests on change. It also makes the tests take longer.

We can add a `coverage` script that is just an alias for `test:unit:coverage` if we want it to be easier to generate the coverage report.

```suggestion
    "test": "npm run lint && npm run check && npm run test:unit",
```

Benaiah (changes requested):
One small change to request but this looks good otherwise!

playwright.config.ts
```diff
@@ -1,11 +1,26 @@
...
+    : [["list"], ["html", { outputFolder: "e2e-tests/report", open: "on-failure" }]],
+  use: {
+    trace: "on-first-retry",
+  },
   webServer: {
     command: "npm run build && npm run preview",
     port: 4173,
+    reuseExistingServer: !process.env.CI,
+    // Set a ridiculously long timeout to allow the site to be built.
+    timeout: 60 * 60 * 1000,
```
hinzed1127, line 21: if we ever hit this, something has gone terribly wrong

hinzed1127 (approved):
Looks good after @Benaiah 's suggested command tweak!

package.json
LouisFettet: Right, I forgot that I added `run` to the command to ensure we weren't rerunning coverage with the watch mode (since the coverage would only update based on the tests that were run, which was misleading). I'll rever that change.

Benaiah (approved):

LouisFettet:
End-to-end testing running on `main` after merge:
Screenshot 2023-04-26 at 5 06 01 PM

LouisFettet:
Also set up required status checks:
Screenshot 2023-04-26 at 5 10 31 PM

PR #62, opened 2023-04-22
Allow building the site without manually creating a .env file

Benaiah:
## Proposed changes

This lets anyone `git clone` the project, run `npm install`, and then run `npm run build`, and have it all work as expected without any additional configuration.

- Adds environment variables required for `/.build-info` to build to `.env.example`
- Adds a `scripts/pre-build.js` script that runs before the build runs
- Makes `scripts/pre-build.js`, when running on a developer machine (i.e., not in Vercel), copy `.env.example` to `.env` if `.env` doesn't already exist.


LouisFettet (approved):
I was considering doing something similar in #59, since Playwright requires `npm run build` to work before it can run tests, and the build was failing due to lack of environment variables. For that use-case I was checking in a `.env.ci` file with blank values which Vite would pick up by passing the `--mode ci` flag, but I like this approach better.

PR #63, opened 2023-04-22
Add Button component

Benaiah:
Jira ticket: [LDAF-176](https://ldaf.atlassian.net/browse/LDAF-176)

## Proposed changes

- Adds a `Button` component with variants based on the LDAF design system.

## Screenshots

image

[LDAF-176]: https://ldaf.atlassian.net/browse/LDAF-176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/stories/Button.stories.ts
```diff
@@ -0,0 +1,79 @@
...
+    variant: {
+      control: { type: "select" },
+      options: variants,
+    },
+    type: {
+      control: { type: "select" },
+      options: types,
+    },
+  },
+} satisfies Meta<Button>;
```
LouisFettet, line 20: We should link to the Figma file here as well! I think this should work:
```js
  parameters: {
    design: {
      type: "figma",
      url: " https://www.figma.com/file/oGKbyCnCRRdNzLYbiags93/LDAF-Component-Library-USWDS-3.0.2?node-id=2196-3764&t=HICajhP8FIexorTH-4",
    }
  }
```

src/lib/components/Button/Button.scss
```diff
@@ -0,0 +1,160 @@
...
+    background: map.get($colors, "primary-darker");
+  }
+  &:disabled {
+    background: map.get($colors, "disabled");
+  }
+}
+
+.usa-button--big-inverse,
+.usa-button.usa-button--big-inverse {
+  @extend .usa-button--big;
```
LouisFettet, line 145: This doesn't seem to be working (at least not in the story):
Screenshot 2023-04-26 at 2 41 12 PM

src/lib/components/Button/Button.svelte
```diff
@@ -0,0 +1,44 @@
...
+
+  let className = "";
+  export { className as class };
+</script>
+
+<button
+  {type}
+  {disabled}
+  aria-disabled={disabled}
+  class={classNames("usa-button", unstyled && "usa-button--unstyled", ...variantClasses, className)}
```
LouisFettet, line 40: Might be better to move the result of `classNames()` into a reactive value; I'm seeing some oddities in Storybook when switching around the values of different props (can't repro anything consistently unfortunately, and I'm seeming to have different results between running it locally versus building and serving it).

I think it would also be useful to add a test around ensuring that the expected classes get applied given different variants.

src/lib/components/Button/__tests__/ButtonTest.svelte
LouisFettet, line 1: Is it possible to use `src/lib/components/__tests__/SlotWrapper.svelte` for this instead? Or is it that the `on:click` handler can't be passed as part of `...$restProps`?

LouisFettet (commented):
Seeing some weirdness in Storybook and I think we can improve testing a bit, but overall this is looking good!

src/lib/components/Button/Button.scss
```diff
@@ -0,0 +1,155 @@
...
+  "base-lightest": #f2f1f0,
+  "base-light": #a9aeb1,
+  "base-dark": #565c65,
+  "base-darker": #3d4551,
+  "gray-05": #f0f0f0,
+  "primary-lightest": #d1e9ff,
+  "text-only-hover": rgba(6, 60, 122, 16%),
+  "text-only-active": rgba(6, 60, 122, 30%),
+  "outline-inverse": #dcdee0,
+);
```
hinzed1127, line 18: My site title PR defined the current color palette as [global variables](https://github.com/la-ldaf/ldaf-site/pull/50/files#diff-0a91e0c8de5f30d8b403617ebcd0339692b61f0ce8b5cb4106b5214ba6058842). That covers everything here except for the last 3 (and `gray-05`, although `$grayscale-05` is extremely close if that's an acceptable substitution, cc @getpunched )

If we like that approach, I can make a smaller PR to just get that in or do follow-up on this once #50 gets in.

hinzed1127 (approved):
Looks good! Up to you how we handle order of operations on the scss variables

src/lib/components/Button/Button.scss
Benaiah: A follow-up PR to add that would be great! I've added a TODO comment indicating that we should do that.

PR #74, opened 2023-04-25
Make Storybook preview import global app styles

Benaiah:
## Proposed changes

- Imports `src/app.scss` in `.storybook/preview.ts`


LouisFettet (approved):
I'm pretty sure we had discussed doing this in a separate PR and I can't remember why we didn't do it? Regardless this is good by me.

PR #75, opened 2023-04-26
Add lazy image loading Storybook example

Benaiah:
## Proposed changes

- Organizes Storybook stories into folders
- Adds a demo of `IntersectionObserver` based lazy image loading to Storybook

## Screenshots

image

### DIFF
--- a/src/lib/components/Image/types.ts
+++ b/src/lib/components/Image/types.ts
@@ -1,1 +1,3 @@
 export type Color = { r: number; g: number; b: number };
+export type Loading = "eager" | "lazy";
+export type LazyLoading = "none" | "native" | "intersectionObserver";
--- a/src/stories/Examples/LazyImageLoading/LazyImageLoading.stories.ts
+++ b/src/stories/Examples/LazyImageLoading/LazyImageLoading.stories.ts
@@ -1,0 +1,15 @@
+import LazyImageLoading from "./LazyImageLoading.svelte";
+import BlurhashRendererDecorator from "../../decorators/BlurhashRendererDecorator.svelte";
+import type { Meta, StoryObj } from "@storybook/svelte";
+
+const meta = {
+  title: "Examples/Lazy Image Loading",
+  component: LazyImageLoading,
+  // @ts-expect-error Unclear how to resolve SvelteComponent <=> SvelteStoryResult type mismatch.
+  decorators: [() => BlurhashRendererDecorator],
+} satisfies Meta<LazyImageLoading>;
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+export const Example: Story = {};
--- a/src/stories/Examples/LazyImageLoading/LazyImageLoading.stories.ts
+++ b/src/stories/Examples/LazyImageLoading/LazyImageLoading.stories.ts
@@ -7,6 +7,12 @@ const meta = {
   component: LazyImageLoading,
   // @ts-expect-error Unclear how to resolve SvelteComponent <=> SvelteStoryResult type mismatch.
   decorators: [() => BlurhashRendererDecorator],
+  argTypes: {
+    lazyLoadingType: {
+      control: { type: "radio" },
+      options: ["none", "native", "intersectionObserver"],
+    },
+  },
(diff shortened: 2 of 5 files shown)


src/stories/Examples/LazyImageLoading/LazyImageLoading.svelte
```diff
@@ -0,0 +1,93 @@
...
+
+  import sampleImage from "../../../sample.jpg";
+  import sampleImageBlurhash, {
+    width as sampleImageWidth,
+    height as sampleImageHeight,
+    mean as sampleImageMean,
+  } from "../../../sample.jpg?blurhash";
+
+  export let nativeLazyLoading = false;
+  export let intersectionObserverLazyLoading = true;
```
LouisFettet, line 13: I keep getting tripped up on the interaction between these two; only one of them gets applied in the end (either we let the browser handle it natively or we use the intersection observer), right? Would it be possible to add comments here that would show up in the story? Don't need a full logic table; just a "hey, the intersection observer doesn't do anything if we're using the native lazy loading" (assuming that's correct).

Alternatively, if the story isn't useful for anything other than showing the intersection observer, let's just disable the controls in the story file:
```js
parameters: {
  controls: {
    disable: true,
  }
}
```

Benaiah: Good point on how the two booleans make it hard to see what's going on - I've refactored `Image` and `LazyImageLoading` to take an explicit lazy image loading type so we can debug it directly without remembering the right incantation of boolean settings for the thing we want to test. Can you take another look at it and let me know what you think of it now?

src/stories/Examples/LazyImageLoading/LazyImageLoading.stories.ts
```diff
@@ -0,0 +1,21 @@
+import LazyImageLoading from "./LazyImageLoading.svelte";
+import BlurhashRendererDecorator from "../../decorators/BlurhashRendererDecorator.svelte";
+import type { Meta, StoryObj } from "@storybook/svelte";
+
+const meta = {
```
hinzed1127, line 5: It could be helpful to add a brief [description](https://storybook.js.org/docs/react/api/doc-block-description#writing-descriptions) via the `parameters.docs.description.component` property

hinzed1127 (approved):
Functionally looks good, just a small documentation request

src/stories/Examples/LazyImageLoading/LazyImageLoading.stories.ts
Benaiah: I've added a full document explaining how to use the example.

src/lib/components/Image/types.ts
```diff
@@ -1 +1,3 @@
 export type Color = { r: number; g: number; b: number };
+export type Loading = "eager" | "lazy";
+export type LazyLoading = "none" | "native" | "intersectionObserver";
```
LouisFettet, line 3: This is all much clearer now, thanks!

LouisFettet (commented):
This looks great! I'm having trouble running `npm run storybook` though (building and previewing works fine); are you seeing something similar?

Otherwise I have just a minor comment on the docs, but this is all really great work!

Benaiah:
> I'm having trouble running npm run storybook though (building and previewing works fine); are you seeing something similar?

@LouisFettet I'm not having any trouble running storybook on this PR, what are you seeing?

LouisFettet:
Getting the following error; I thought that nuking my `node_modules` might fix it but no dice:
```
Failed to fetch dynamically imported module: http://localhost:6006/node_modules/.cache/.vite-storybook/deps/@storybook_svelte_preview.js?v=878f166b
```

Not getting the same issue on `main` so I assume there's either something up with my machine or we're missing something here. Which is confusing because this works fine when running a full storybook build.

LouisFettet (approved):
Discussed the problem [in Slack](https://adhoc.slack.com/archives/C04EYUA76BT/p1682617801745479?thread_ts=1682617539.141839&cid=C04EYUA76BT) and the issue was resolved after pulling in latest dependency updates! :shipit:

PR #79, opened 2023-04-26
Button: fix unstyled button :hover and :active styles

Benaiah:
Followup to #63.

## Proposed changes

- Adds explicit background and foreground color settings to `.usa-button.usa-button--unstyled`.

### DIFF
--- a/src/lib/components/Button/Button.scss
+++ b/src/lib/components/Button/Button.scss
@@ -79,46 +79,28 @@ $colors: (
     background: none;
     color: map.get($colors, "disabled");
   }
-}
-
-.usa-button.usa-button--outline {
-  background: none;
-  color: map.get($colors, "primary");
-  border: 2px solid map.get($colors, "primary");
-  box-shadow: none;
-  &:hover {
+  &:disabled:hover,
+  &:disabled:active {
     background: none;
-    color: map.get($colors, "primary-dark");
-    border: 2px solid map.get($colors, "primary-dark");
-    box-shadow: none;
-  }
-  &:active {
-    background: none;
-    color: map.get($colors, "base-darker");
-    border: 2px solid map.get($colors, "base-darker");
-    box-shadow: none;
-  }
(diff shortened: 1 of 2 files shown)


LouisFettet (commented):
I think that the `outline` and `outline-inverse` variants are still affecting the button when it's `unstyled`, which I assume is not intentional.

Benaiah:
> I think that the `outline` and `outline-inverse` variants are still affecting the button when it's `unstyled`, which I assume is not intentional.

That's what I thought too, but the [USWDS code sample for the Button component](https://designsystem.digital.gov/components/button/) actually shows that the `usa-button--unstyled` class is supposed to work on _all_ the button variants.

LouisFettet:
This is what I'm seeing in storybook (similar for `outline-inverse` and when it's disabled):
Screenshot 2023-04-27 at 12 31 40 PM

Whereas in the USWDS docs the unstyled button with the "outline" variant doesn't show that border.

Benaiah:
Oh, my bad! I thought I'd already fixed that. Should be fixed now, and I've also fixed the color of the unstyled inverse button.

LouisFettet:
Sorry, a couple more combinations I'm seeing with oddities:
1. `text-only` variant that's `disabled` has contrast issues with hover and active states:
Screenshot 2023-04-27 at 1 10 17 PM

2. `outline` and `outline-inverse` variants that are `disabled` _sometimes_ have some sizing issues on hover and active states:

Kapture 2023-04-27 at 14 07 11

Benaiah:
@LouisFettet I've rebased this and (I _think_) fixed all the style issues with every permutation of `variant`, `disabled`, and `unstyled`. Thanks for the thorough review! It's ready for another look now.

LouisFettet (approved):
Looks great! Thanks for dealing with all my comments. After we merge this and #50 we'll want to do some follow-up work to standardize how we're handling colors.

PR #103, opened 2023-05-04
LDAF-249 Set up basic Contentful fetch service and example page

LouisFettet:
Jira ticket: [LDAF-249](https://ldaf.atlassian.net/browse/LDAF-249)

Gets some basic scaffolding in place for pulling content from Contentful and shows how we can handle building pages when a Contentful fetch fails (use-case now is for folks without access to our secrets, but theoretically we could so something similar if there's some sort of outage).

## Proposed changes

- Adds environment variables to `.env.example` for the Vercel flag and our Contentful instance's space ID and delivery and preview API tokens (you can get these locally using the Vercel CLI: `vercel env pull`).
- Sets up a very basic Contentful service which allows for fetching via the Contentful GraphQL API.
  - Handles some checks of required environment variables before attempting requests.
  - We may end up replacing this with an SDK, so I held off on trying to set up any GraphQL type defs.
- Updates the `/test-contentful-content` page with real "fake" content from Contentful (see "Test Contentful Content Page" in Contentful). If the request fails for any reason, it falls back to the test Markdown doc (and if that fails we render an error).
- Minor updates to the `ContentfulRichText` component:
  - Moves some of the Markdown logic into the route that's rendering the component.
  - Have it accept a content prop (an array of blocks) rather than a document prop; we might want to revert this later if there's document metadata we want to utilize in the component.
- Some miscellaneous smaller updates:
  - Sets up a close-to-default `+error.svelte` page which uses the right layout and can accept a title (an error message was already required).
  - Adds `bundle-visualizer` to our ignore files for ESLint and Prettier.

- [x] Set up relevant API tokens.
- [x] Ensure site builds normally with and without Contentful API tokens.

## Requested feedback

See self-review; otherwise I think "done is better than perfect" here, so it might be good to just get this through so we can use some of this to start working with the SDK and getting previewing working. 

[LDAF-249]: https://ldaf.atlassian.net/browse/LDAF-249?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app.d.ts
```diff
@@ -2,7 +2,10 @@
 // for information about these interfaces
 declare global {
   namespace App {
-    // interface Error {}
+    interface Error {
+      // `message` is included by default
+      title?: string;
```
LouisFettet, line 7: We'll default to the error code if a title isn't provided, but I figured it would be good to start stubbing out other fields here we could use for error messaging.

src/lib/components/ContentfulRichText/ContentfulRichText.svelte
```diff
@@ -1,16 +1,10 @@
...
-  import type { Document } from "@contentful/rich-text-types";
-  import { markdownDocument } from "./__tests__/documents";
-  import { isDocument } from "./predicates";
+  import type { TopLevelBlock } from "@contentful/rich-text-types";
 
-  export let document: Document = markdownDocument.document;
-
-  if (!isDocument(document)) {
-    throw new Error("document prop is not a document");
-  }
```
LouisFettet, line 11: Moved this logic into `src/routes/test-contentful-content/+page.server.ts`, but we may want to move to some sort of utility function in the future.

src/lib/components/ContentfulRichText/ContentfulRichText.test.ts
```diff
@@ -1,14 +1,14 @@
 import "@testing-library/jest-dom";
 import { render } from "@testing-library/svelte";
 import { describe, it, expect } from "vitest";
-import "$lib/__tests__/extendExpect.ts";
+import "$lib/__tests__/extendExpect";
```
LouisFettet, line 4: Removing the `.ts` resolved a `rollup` warning we were getting in the build logs.

src/lib/services/contentful/fetch.test.ts
```diff
@@ -0,0 +1,59 @@
...
+
+vi.mock("$env/static/private");
+type EnvVarName = "VERCEL" | "CONTENTFUL_SPACE_ID" | "CONTENTFUL_DELIVERY_API_TOKEN";
+type EnvVars = Record<EnvVarName, string>;
+const setEnvVars = (vars: EnvVars) => {
+  let key: keyof EnvVars;
+  for (key in vars) {
+    (envVars as EnvVars)[key] = vars[key];
+  }
+};
```
LouisFettet, line 13: After much floundering with getting a mock working here, I ultimately landed here after shamelessly stealing from @Benaiah's work on the Intersection Observer test: https://github.com/la-ldaf/ldaf-site/blob/1e67dd643afd6accf014629c68cc3696a43416fc/src/lib/components/IntersectionObserver/IntersectionObserver.test.ts#L12-L18

There really doesn't seem to be a great way to mock `$env/static/private` (or any of the other SvelteKit `$env/*` imports), and since `vi.mock` gets hoisted I couldn't really find a cleaner way of doing this. Some other attempts and considerations included:
* Tried using `vi.stubEnv`, which seemed like it would have been perfect, except by the time the tests run it's too late to update env vars (changes don't get reflected in `$env/*`)
* Tried using `vi.doMock` within each test, which didn't work and also didn't seem like the right use-case since we're mocking an import of the file we want to test and not the file itself.
* I would have liked to use something like `vi.fn().mockImplementationOnce()` but `$env/*` doesn't utilize functions for me to mock.
* Considered setting up a different mock of `$env/static/private` in `__mocks__/` for each use-case, which just felt a bit messy, so I didn't attempt it (not even sure if it would have worked).

All that to say, I'm still not super happy with this, so I'm very open to any suggestions on how we can improve it. It does work as-is though, so I'm also fine with moving forward with what we've got.

src/lib/services/contentful/fetch.ts
```diff
@@ -0,0 +1,31 @@
...
+const graphApiUrl = (): string =>
+  `https://graphql.contentful.com/content/v1/spaces/${CONTENTFUL_SPACE_ID}`;
+
+const graphApiOptions = () => ({
+  method: "POST",
+  headers: {
+    "Content-Type": "application/json",
+    Authorization: `Bearer ${CONTENTFUL_DELIVERY_API_TOKEN}`,
+  },
+});
```
LouisFettet, line 17: Obviously I'd prefer to evaluate these values only once, instead of on every request. If you've got some ideas let me know!

LouisFettet (commented):
Self-review!

src/lib/services/contentful/fetch.ts
```diff
@@ -0,0 +1,31 @@
+import { VERCEL, CONTENTFUL_SPACE_ID, CONTENTFUL_DELIVERY_API_TOKEN } from "$env/static/private";
+
+// Re-evaluate utils using env vars on each request to ensure tests can
+//   properly mock and overwrite them.
+const contentfulConnected = (): boolean =>
+  !!(VERCEL === "1" && CONTENTFUL_SPACE_ID && CONTENTFUL_DELIVERY_API_TOKEN);
```
hinzed1127, line 6: you can basically infer the meaning of `VERCEL`, but maybe it's worth documenting, probably in the `.env` file, the [documented definition](https://vercel.com/docs/concepts/projects/environment-variables/system-environment-variables#system-environment-variables) of it

src/lib/services/contentful/fetch.ts
hinzed1127: two thoughts:

1. I might be missing something obvious, but is there a reason these can't just be declared as objects instead of functions?
2. If (1) isn't an option, would this be a valid use of a [`readable` store](https://svelte.dev/docs#run-time-svelte-store-readable)?

src/lib/services/contentful/fetch.test.ts
hinzed1127: I am never above doing whatever witchcraft is required to make test mocks work as desired 😅

hinzed1127 (approved):
very on board with "done is better than perfect". this all looks solid to me!

hinzed1127 (approved):
very on board with "done is better than perfect". this all looks solid to me!

src/lib/services/contentful/fetch.ts
LouisFettet: I think using objects would still give us trouble when testing; the readable store might work with a mock though. I had made one early attempt at using stores here (never committed), so I could give that another go. Will give that a try after I finish up some things with the nav in #132.

src/lib/services/contentful/fetch.ts
LouisFettet: Yeah the idea of using `VERCEL` here was that a contributor wouldn't have the Contentful id and key without also being able to run `vercel env pull` (so even though the doc says that "the app is deployed and running on Vercel" it's also `1` for us locally), but it's rather redundant since I still want to ensure that the Contentful id and key are defined. Will probably adjust this to remove the usage of `VERCEL`.

src/lib/services/contentful/fetch.ts
Benaiah: I'm going to take a shot at representing these without functions, I think we can manage it in a similar way to how we mock the env variables.

src/lib/services/contentful/fetch.test.ts
Benaiah: I'm fine with this approach! Looks similar to what I did in my intersection observer tests. I think we might be able to re-use this technique for the other file below, going to give this a shot.

Benaiah (commented):
Looks good for the most part! Small nit about removing the `isDocument` check from the component and changing the `document` prop to `contents` (I think we can leave both of those as they were), and I'm going to try to fix the functions-as-constants bit in `src/lib/services/contentful/fetch.ts`, but other than that this looks good to go!

src/lib/components/ContentfulRichText/ContentfulRichText.svelte
LouisFettet: Moved things back in f22c49c.

src/lib/services/contentful/fetch.test.ts
LouisFettet: @Benaiah sounds good! To get this moving along, are you okay with trying that out in a follow-up, or did you want to see about getting that change in here?

src/lib/services/contentful/fetch.ts
LouisFettet: Updated to remove redundancy in f652535!

src/routes/test-contentful-content/+page.svelte
```diff
@@ -1,5 +1,7 @@
 <script lang="ts">
```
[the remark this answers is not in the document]
LouisFettet: I don't think that this type declaration was necessary to get things working (pretty sure this is inferred properly), but if I have issues in the follow-up I might use this.

LouisFettet:
@Benaiah I made a few updates if you want to give this another look!

src/routes/test-contentful-content/+page.svelte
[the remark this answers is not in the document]
Benaiah: Weird, it was showing up as `any` in my IDE without this change.

src/lib/services/contentful/fetch.test.ts
Benaiah: I'm fine with trying that out in a follow-up and approving this to go in for now, sure!

src/lib/services/contentful/fetch.test.ts
```diff
@@ -0,0 +1,57 @@
...
+
+describe("Contentful Fetch", () => {
+  beforeEach(() => {
+    // Re-initialize fetch mock every time (since we're clearing it every time).
+    global.fetch = vi.fn(async () => Promise.resolve(true)) as Mock;
+  });
+
+  afterEach(() => {
+    // Clear mock history and reset implementations to empty functions.
+    vi.resetAllMocks();
```
Benaiah, line 25: Nit: if you use `vi.restoreAllMocks()` here instead of `vi.resetAllMocks()`, you can initialize `fetch` once instead of in a `beforeEach`. Either way works tho.

Benaiah (approved):
Looks great, thanks for all the work!

src/lib/services/contentful/fetch.test.ts
LouisFettet: Ah you're right, this was leftover from a previous implementation where I was mocking the env vars a different way. I think we can actually just mock fetch once at the top and use `vi.clearAllMocks()` as long as we're going to call `setEnvVars` at the top of each test case.

I'll update this and make a note of it.

src/routes/test-contentful-content/+page.svelte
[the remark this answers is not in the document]
LouisFettet: Oh odd, if this is still happening after merge I can address this in #132.

PR #110, opened 2023-05-05
Add support for Vercel Speed Insights API

### LINKED ISSUE #106: Implement support for Vercel's Speed Insights API
Jira ticket: [LDAF-299](https://ldaf.atlassian.net/browse/LDAF-299)

## Problem statement

Implement support for Vercel's [Speed Insights API](https://vercel.com/docs/concepts/speed-insights/api) to get page performance measurements from real users.

## Background

TODO

## Acceptance criteria

- [ ] TODO

TODO (see https://github.com/vercel/vercel/blob/main/examples/sveltekit/src/lib/vitals.js)

[LDAF-299]: https://ldaf.atlassian.net/browse/LDAF-299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Benaiah:
## Closes #106 

Jira ticket: [LDAF-299](https://ldaf.atlassian.net/browse/LDAF-299)

## Proposed changes

- Adds support for Vercel's Speed Insights API (code adapted from [Vercel's SvelteKit Speed Insights example](https://github.com/vercel/vercel/tree/main/examples/sveltekit).

I don't think there's much we can do to validate that this is working properly until we merge it and confirm that data is being sent to Vercel.

[LDAF-299]: https://ldaf.atlassian.net/browse/LDAF-299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/routes/+layout.svelte
```diff
@@ -1,18 +1,31 @@
...
   import { BlurhashRenderer } from "$lib/components/Image";
   import { key as pageMetadataMapKey } from "$lib/context/pageMetadataMap";
 
+  $: if (browser && analyticsID) {
+    webVitals({
+      path: $page.url.pathname,
+      params: $page.params,
+      analyticsID,
+    });
+  }
```
LouisFettet, line 27: We'll want to check that this isn't firing more than once after we deploy.

LouisFettet (approved):
Curious to see how this works once we deploy! I wonder if some metrics like TTFB will work if we're only loading the metrics after the user has started loading the page?

Regardless, I don't see any downsides to this update so let's :shipit: and see how it goes!

Benaiah:
> I wonder if some metrics like TTFB will work if we're only loading the metrics after the user has started loading the page?

See https://github.com/GoogleChrome/web-vitals#install-and-load-the-library, where it says the following:

> The web-vitals library uses the buffered flag for [PerformanceObserver](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver/observe), allowing it to access performance entries that occurred before the library was loaded.

src/routes/+layout.svelte
Benaiah: This will fire whenever `$page` changes, so client-side routing will also be included. This seems to be intended because it matches the behavior of Vercel's example, but we can always disable it if it's not working the way we want.

PR #127, opened 2023-05-10
Add Card Component

hinzed1127:
## Closes #12345 

Jira ticket: [LDAF-218](https://ldaf.atlassian.net/browse/LDAF-218)

## Proposed changes

- Adds Card component with variations designed [here](https://www.figma.com/file/oGKbyCnCRRdNzLYbiags93/LDAF-Component-Library-USWDS-3.0.2?type=design&node-id=2196-4072&t=FuAC5ahVOuORPZL4-0)

## Screenshots
primary card
secondary card
task list card

- [ ]  Matches design
- [ ]  Added variation examples on the `/news` sub-route to show how it's actually used outside of storybook

### Possible drawbacks

- This was more difficult to add into storybook, because as it stands, [named slots can't be wrapped in a condition](https://github.com/sveltejs/svelte/issues/5604). Given that the component's design doesn't necessarily suffer from this, just the tooling around it (i.e., testing and displaying in storybook), I think this is an acceptable thing to deal with

## Requested feedback

- what do y'all think about the named slots? It keeps the component slimmer from an arguments perspective, but wanted to make sure it doesn't create deficiencies elsewhere

[LDAF-218]: https://ldaf.atlassian.net/browse/LDAF-218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/stories/CardView.svelte
```diff
@@ -0,0 +1,78 @@
...
+            <Image
+              src={sampleImage}
+              alt="Primary view placeholder"
+              width={imageWidth}
+              height={imageHeight}
+              mean={imageMean}
+              blurhash={imageBlurhash}
+            />
+          {/if}
+        </div>
```
LouisFettet, line 41: nit: I noticed that the following line doesn't seem to be getting covered via tests: https://github.com/la-ldaf/ldaf-site/blob/b7bd4afbdd2763cdd4e0ff08e3b8d1af27f278ac/src/lib/components/Card/Card.svelte#L11

Maybe we should move the `#if` check one line higher to wrap the `slot="image"` container to make sure that we apply the classes properly? Disregard if this messes up the layout or something!

src/stories/Card.stories.ts
```diff
@@ -0,0 +1,61 @@
...
+  argTypes: {
+    variant: {
+      control: { type: "select" },
+      options: variants,
+    },
+  },
+  parameters: {
+    design: {
+      type: "figma",
+      url: "https://www.figma./file/oGKbyCnCRRdNzLYbiags93/LDAF-Component-Library-USWDS-3.0.2?node-id=2196-4072&t=vo32NlfE5S0IJB6g-0",
```
LouisFettet, line 20: I think this URL is malformed (missing the `.com`).

src/lib/components/Card/Card.svelte
```diff
@@ -0,0 +1,59 @@
...
+    secondary: "",
+  };
+  $: variantClass = variantClassesDict[variant];
+
+  let className = "";
+  export { className as class };
+  $: classes = classNames("usa-card", variantClass, className);
+</script>
+
+<li class={`${classes}`}>
```
LouisFettet, line 21: nit: this can just be simplified to this right?
```suggestion
<li class={classes}>
```

src/lib/components/Card/Card.svelte
```diff
@@ -0,0 +1,59 @@
...
+  <div class="usa-card__container">
+    {#if $$slots.header}
+      <div class="usa-card__header">
+        <slot name="header">
+          <h2 class="usa-card__heading">Default header</h2>
```
LouisFettet, line 26: Do we need these defaults if we're always checking that a slot is provided before rendering? Or is the point here to show an example of the slot content?

LouisFettet (commented):
This looks great, and I really like the usage of named slots here! Just have a handful of mostly minor comments for you.

src/lib/components/Card/Card.test.ts
```diff
@@ -0,0 +1,40 @@
+import "@testing-library/jest-dom";
+import { waitFor, render, screen } from "@testing-library/svelte";
+import { describe, it, expect } from "vitest";
+
+import CardView from "../../../stories/CardView.svelte";
```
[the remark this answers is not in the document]
hinzed1127: yeah I'm fine with a `TODO` for now.

The only other idea I have re options is to create a `components/util` directory or something to that effect, where that's the designated spot for wrapper components that are need for tests and/or stories, which isn't a huge improvement, but could be a way to consolidate these things effectively

src/lib/components/Card/Card.svelte
hinzed1127: it sure can 🤦

src/lib/components/Card/Card.svelte
hinzed1127: that's a good catch. I think this was before I started utilizing the `$$slots` argument, so this default will never actually render in the current code.

src/lib/components/Card/cardOptions.ts
```diff
@@ -0,0 +1,3 @@
+export const variants = ["primary", "secondary"] as const;
+
+export type Variant = (typeof variants)[number];
```
hinzed1127, line 3: I initially had this when I assumed there would be more variations dependent on classes, but with the named slots and the only image version of the card being the flag orientation, I think we can remove this entirely for now.

src/stories/CardView.svelte
hinzed1127: Alrighty, it's not the prettiest way of handling it, but since it's only within Storybook, I figure this is an acceptable way to toggle at least the named image and footer slots in 8b16d30d2b6b667d58d954e9560e0771fde820ec and fa64fd0a961192bf45a82f3ef326d6da3e0d3068, respectively

src/lib/components/Card/Card.svelte
```diff
@@ -1,58 +1,38 @@
...
 
-<li class={`${classes}`}>
+<li class={classes}>
   <div class="usa-card__container">
     {#if $$slots.header}
       <div class="usa-card__header">
-        <slot name="header">
-          <h2 class="usa-card__heading">Default header</h2>
-        </slot>
+        <!-- Use the appropriate section heading element, likely <h2> or <h3>, based on page structure -->
```
LouisFettet, line 14: Appreciate these comments you added!

LouisFettet (approved):
Latest changes all look good to me! Kind of a shame there's not a better way to render things properly in Storybook but I think that your workaround is good to get things looking consistent. Thanks for all the work here!

Benaiah (approved):

src/lib/components/Card/Card.svelte
hinzed1127: Yeah, since there's not a way to directly provide descriptions of a named slot in storybook, and since to my knowledge, you could conceivably put any element or component within a slot, this was the best method I landed on to make things slightly self-documenting

PR #131, opened 2023-05-11
Add readme to vite-plugin-import-as-bundle-string

Benaiah:
## Proposed changes

- Adds readme to `vite-plugin-import-as-bundle-string`


LouisFettet (approved):
LGTM! Thanks for adding all these docs; super helpful!

PR #150, opened 2023-05-18
Generate GraphQL schema

Benaiah:
Depends on #132.

## Proposed changes

- Generates the GraphQL schema in addition to the TypeScript schema
- Updates both schemas


LouisFettet (approved):
This LGTM after comparing directly to the latest changes on `main`. Looks like we just need to resolve merge conficts (and maybe run `gen-schema` again) and this should be good to go!

PR #151, opened 2023-05-18
ESLint linting for GraphQL queries

Benaiah:
Depends on #150.

## Proposed changes

- Adds and configures `@graphql-eslint/eslint-plugin` to lint GraphQL queries declared with the `gql` tag or in `.graphql` or `.gql` files using the checked-in Contentful schema.
- Adds `graphql` and `graphql-tag`; uses `gql` on the server side to generate GraphQL queries that are linted by the ESLint plugin and then uses `graphql`'s `print` function to turn them back into strings (we'll do more interesting things with the parsed queries in the future).
- Names the GraphQL queries to get rid of a lint warning.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -11,6 +11,8 @@ node_modules
 /e2e-tests/report
 /e2e-tests/results
 /bundle-visualizer
+src/lib/services/contentful/schema.ts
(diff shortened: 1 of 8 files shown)


LouisFettet (approved):
LGTM! This is a great improvement!

PR #161, opened 2023-05-22
Move Meta-Site from GitHub Pages to Vercel

LouisFettet:
Jira ticket: [LDAF-321](https://ldaf.atlassian.net/browse/LDAF-321)

In order to more efficiently build and deploy Storybook, we should use Vercel over GitHub Pages for the Meta-Site.

## Proposed changes

- Replace deploy to GitHub Pages in GitHub Actions workflow with a deploy to Vercel (preview for commits to PRs and production for commits to `main`): https://vercel.com/ldaf/meta-site
- Check the Meta-Site landing page into the source code under `.meta/` and have it render with slight variations for different cases (e.g. show PR info for previews): https://meta-site-161-ldaf.vercel.app/
- Update the `/.build-info` route to point to the Meta-Site in Vercel rather than GitHub Pages: https://ldaf-git-louis-vercel-meta-site-ldaf.vercel.app/.build-info

- [x] Move the Meta-Site to Vercel.
- [x] Build a preview of the Meta-Site on every commit to a PR and on every commit to the main branch.

## Requested feedback

No specific feedback requested, but I left a self-review for some additional context.

[LDAF-321]: https://ldaf.atlassian.net/browse/LDAF-321?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/tests_and_reports.yml
```diff
@@ -52,7 +49,7 @@ jobs:
         run: npm run test:unit:coverage
 
       - name: Report code coverage to PR in comment
-        if: github.ref != 'refs/heads/main'
+        if: ${{ !fromJSON(env.IS_MAIN) }}
```
LouisFettet, line 52: I was hoping to just run `!env.IS_MAIN` but the value of `IS_MAIN` is a string, not a boolean, so we need to use the `fromJSON` helper to cast it to a boolean. Not super happy with this, but it's probably still better than checking if `github.ref == 'refs/heads/main'` every time.

.github/workflows/tests_and_reports.yml
```diff
@@ -126,59 +123,73 @@ jobs:
...
+  deploy-reports-to-vercel-meta-site:
+    name: "Deploy Reports to Vercel Meta-Site"
     # Wait for tests to complete before deploying.
     needs: [lint-and-unit-tests, end-to-end-tests, bundle-visualizer]
     # We still want to upload reports regardless of whether tests failed.
     if: ${{ always() }}
     runs-on: ubuntu-latest
+    env:
+      VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
+      VERCEL_PROJECT_ID: ${{ secrets.VERCEL_META_SITE_PROJECT_ID }}
```
LouisFettet, line 136: These env vars just need to be set; the Vercel CLI will automatically pick these up, so we don't need to provide them as flags (but for some reason we do need to provide `VERCEL_TOKEN` each time).

.github/workflows/tests_and_reports.yml
```diff
@@ -126,59 +123,75 @@ jobs:
...
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          publish_branch: gh-pages
-          publish_dir: ./pr-reports
-          destination_dir: ${{ github.ref == 'refs/heads/main' && env.MAIN_PATH || env.PR_PATH }}
+      - name: Deploy to Vercel preview
+        if: ${{ !fromJSON(env.IS_MAIN) }}
+        # Guide: https://vercel.com/guides/how-to-alias-a-preview-deployment-using-the-cli
+        run: |
+          url="$(vercel deploy --cwd .meta --prebuilt --token=${{env.VERCEL_TOKEN}})"
+          vercel alias --cwd .meta --scope ldaf --token=${{env.VERCEL_TOKEN}} set "$url" ${{env.VERCEL_PREVIEW_URL}}
```
LouisFettet, line 181: Unfortunately there's no way to set the URL as part of the `deploy` command, so we need to run `alias` after the fact to set it to a more predictable URL (I chose to use the PR number, since transforming the branch name into a slug would have been a little more complicated). If it would be useful, we could also set up an alias using the commit hash.

.meta/vercel.json
```diff
@@ -0,0 +1,7 @@
+{
+  "buildCommand": null,
+  "devCommand": null,
+  "framework": null,
+  "installCommand": null,
+  "trailingSlash": true
```
LouisFettet, line 6: Without this, if you visit a page without a trailing slash that has assets, such as the unit test coverage report, it will incorrectly try to grab the assets from the root instead of relative, e.g. `/asset.js` instead of `/unit-test-coverage/asset.js`

.meta/index.html
LouisFettet, line 1: This is quite messy since I essentially tried to handle a lot of cases in one page:
1. Default, where no branch or PR is detected, and we just say "reports for this branch or PR"
2. The `main` branch, where we indicate that this is the production deployment and remove the link to the production deployment (since you're already there).
3. A PR, where we at least show the PR number and then provide more details if the GitHub API request is successful.

I think ideally we would generate a lot of this at build time using a simple framework and available environment variables instead of doing things on-the-fly clientside based on the URL, but I figured that would be an enhancement we could make later since I've already probably spent too much time on this 😅

LouisFettet (commented):
Self-review!

Benaiah (approved):

PR #165, opened 2023-05-23
Add Contentful connection via SDK, update home page

hinzed1127:
Jira ticket: [LDAF-200](https://ldaf.atlassian.net/browse/LDAF-200), [LDAF-302](https://ldaf.atlassian.net/browse/LDAF-302?search_id=b4654af9-7d4b-46b7-a644-bce63b38cfa9)

## Proposed changes

- Updates home page to "Under construction" page
- Adds the SDK for [Contentful's Content Delivery API](https://github.com/contentful/contentful.js)
- Text content sourced from Contenful

- [x]  Matches @getpunched 's [mocks](https://www.figma.com/file/JdZep2KGWtsAjqaHrfYoYk/LDAF-Prototypes?type=design&node-id=403-6588&t=RAeS4T6RlwM5jkCm-0)
- [x]  Requests entries from Contentful via `contentful` dependency and renders them accordingly (a "hello world" of sorts) 

### Alternate solutions

I initially tried to add images to contentful and source them from there, as well, but since images aren't currently supported by the rich text component, and since we're still undecided on the exact architecture of how we're handling images, I decided to just include them directly from the code for now.

## Requested feedback

[LDAF-200]: https://ldaf.atlassian.net/browse/LDAF-200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LDAF-302]: https://ldaf.atlassian.net/browse/LDAF-302?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/routes/+page.server.ts
```diff
@@ -0,0 +1,19 @@
+import contentful from "contentful";
+import type { ContentfulClientApi } from "contentful";
+import { CONTENTFUL_SPACE_ID, CONTENTFUL_DELIVERY_API_TOKEN } from "$env/static/private";
+
+// TODO: refactor to a reusable helper function that wraps contenfulClient
```
hinzed1127, line 5: @Benaiah I saw you already had abstracted some of this out in your preview work. I can update to reuse that work here when it's ready

src/routes/+page.svelte
```diff
@@ -1,72 +1,56 @@
...
-        do you do your work? Write in the present tense, and stay brief here. People who are
-        interested can find details on internal pages.
-      </p>
-      <a class="usa-button usa-button--big" href="/">Call to action</a>
+    <div class="grid-col-8 padding-left-4">
+      <ContentfulRichText document={data.body} />
     </div>
   </div>
 </section>
-<GraphicList />
```
hinzed1127, line 51: removing these components from the landing page but keeping them around for the time being

src/routes/page.scss
```diff
@@ -0,0 +1,18 @@
...
+}
+
+.construction-sign-container {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+.homepage-greeting {
+  word-wrap: anywhere;
```
hinzed1127, line 17: this was to deal with the url. I imagine that we're going to avoid raw urls for the most part, but perhaps this is something to add at least to the link component to handle this on smaller widths?

src/routes/+page.server.ts
```diff
@@ -0,0 +1,19 @@
...
+    space: CONTENTFUL_SPACE_ID,
+    accessToken: CONTENTFUL_DELIVERY_API_TOKEN,
+  });
+
+  const title = await contentfulClient.getEntry("5k6saMkIV7PHFhNDxmZiIG");
+  const body = await contentfulClient.getEntry("1VCVunA3sIV5Ch1qGHEKZZ");
+
+  return {
+    title: title.fields.body,
+    body: body.fields.body,
```
hinzed1127, line 17: I'm getting TypeScript errors right now supplying these as arguments to `<ContentfulRichText />` and cannot figure out for the life of me what to update to based on the type definitions in Contentful

hinzed1127 (commented):
tiny self-review

src/routes/+page.server.ts
Benaiah: This is the type you're looking for:

```typescript
import type { Document } from "@contentful/rich-text-types";
```

Then you can use `as` for now to assert the type. We'll move this over to fully-typed GraphQL queries eventually anyway (GraphQL queries will let us grab all the data we need for a page in a single request).

Benaiah: Actually, we can get rid of the `as`! Try something like this:

```typescript
import contentful, { type ContentfulClientApi, type EntryFieldTypes } from "contentful";

// ...

type EntrySkeleton = {
  contentTypeId: "testRichText",
  fields: { body: EntryFieldTypes.RichText }
};

// ...

export const load = async () => {
  // ...
  const title = await contentfulClient.getEntry<EntrySkeleton>("5k6saMkIV7PHFhNDxmZiIG");
  const body = await contentfulClient.getEntry<EntrySkeleton>("1VCVunA3sIV5Ch1qGHEKZZ");

  return {
    title: title.fields.body,
    body: body.fields.body,
  };
};
```

src/routes/+page.server.ts
hinzed1127, line 1: Updated this to follow a similar pattern as `src/routes/test-contentful-content/+page.server.ts` in handling placeholder data when a contentful connection is lacking

LouisFettet (approved):
Looks great to me for the initial setup! I say we :shipit: to get the homepage updated on prod and we can always make more improvements later.

src/routes/page.scss
hinzed1127: Made a TODO and can potentially create a small ticket to capture this somewhere

PR #170, opened 2023-05-26
Add Lighthouse reports to Meta-Site

LouisFettet:
Jira ticket: [LDAF-252](https://ldaf.atlassian.net/browse/LDAF-252)

## Proposed changes

- Adds Lighthouse audits on every commit.
- Adds a landing page to the meta-site that summarizes the reports.
- Deploys the generated reports to the meta-site and links to it from [a PR comment](https://github.com/la-ldaf/ldaf-site/pull/170#issuecomment-1572505533), the [meta-site landing page](https://meta-site-170-ldaf.vercel.app/), and the [`/.build-info` route](https://ldaf-git-louis-ci-lighthouse-ldaf.vercel.app/.build-info).

- [x] generate a lighthouse report on every commit for homepage and documentation page on both mobile and desktop
- [x] store this report somewhere it can be easily accessed 

## Requested feedback

See self-review; I figure when we get around to updating the README we'll want to add some notes about how this works since I wasn't able to get it working with dev dependencies and npm scripts.

Other than that, the only other thing that I can think of is that we _could_ have this act like a test and give us an error if we don't hit a certain threshold on any given score, but I figured that's something we could add in later if we want it.

[LDAF-252]: https://ldaf.atlassian.net/browse/LDAF-252?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/tests_and_reports.yml
```diff
@@ -126,11 +127,51 @@ jobs:
...
+          npm ci
+          npm i -g lighthouse lighthouse-batch
+
+      - name: Generate Lighthouse reports
+        run: |
+          npm run build
+          npm run preview &
+          sleep 5
+          lighthouse-batch --use-global --html --out lighthouse-reports/mobile --file .lighthouse/urls.txt --params "--only-categories=\"performance,accessibility,best-practices,seo\""
+          lighthouse-batch --use-global --html --out lighthouse-reports/desktop --file .lighthouse/urls.txt --params "--only-categories=\"performance,accessibility,best-practices,seo\" --preset=desktop"
```
LouisFettet, line 158: I tried installing these as dev dependencies and writing these commands under scripts in `package.json`, but I was running into all sorts of issues both locally and in CI, so I opted for the first thing that worked, which is what we've got here (global installs and commands written out in the workflow).

.lighthouse/index.html
LouisFettet, line 1: There weren't any good (or at least, not easy-to-find) solutions for creating a UI for summarizing a set of reports, so I threw this together utilizing the handy `summary.json` output of `lighthouse-batch` runs.

LouisFettet (commented):
Self-review!

.github/workflows/tests_and_reports.yml
Benaiah: That's fine by me, it's working!

Benaiah (approved):

PR #196, opened 2023-06-01
Create office pages

LouisFettet:
Jira ticket: [LDAF-322](https://ldaf.atlassian.net/browse/LDAF-322)

## Proposed changes

- Adds a `/about/organization/[slug]` route for rendering office pages. As with most of our routing setup so far, the URL / directory structure may need to change in the future. 
- Adds a `ContactCard` component for rendering mailing addresses, phone numbers, and email addresses.
  - This component uses a new `CopyToClipboard` component which renders a clickable icon that provides a quick success / error message.
- Stubs out the `Breadcrumb` and `SideNav` components, including tests and stories.
- Removes the `main` tag from the base layout so that we can place it where we want it on other pages.
- Fixes some hex codes for success colors that were mismatched in Figma.
- Updates our schema type def file.

## Screenshots

The relevant Figma mocks is [this desktop mock](https://www.figma.com/file/JdZep2KGWtsAjqaHrfYoYk/LDAF-Prototypes?type=design&node-id=603-12330&t=ZtgymRwX7RLS21wj-4) for the office pages.

Preview is up, here's Forestry: https://ldaf-git-louis-office-page-ldaf.vercel.app/about/organization/forestry

- [x] Office pages up at `/about/organization/*`
- [x] Includes contact information

The ticket also mentions figuring out how to place these pages in the nav, but we're going to save that for the nav strategy ticket.

## Requested feedback

See self-review; got a few questions in there for y'all!

[LDAF-322]: https://ldaf.atlassian.net/browse/LDAF-322?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/routes/+layout.svelte
```diff
@@ -36,9 +38,7 @@
   <Banner />
   <div class="usa-overlay" />
   <Header {navItems} {secondaryNavItems} {siteTitle} bind:navMenuExpanded />
-  <main id="main-content">
```
LouisFettet, line 39: We had a few occurences of providing multiple `main` tags on some pages, so I opted to remove this and leave it up to each route's layout / page to handle adding the tag where appropriate (some need it to wrap everything, while others that have breadcrumbs and sidenavs might want it only wrapping the actual content of the page). This change obviously means that it looks like there's a lot more shifting here than there actually is; I'd suggest viewing the route files [with whitespace changes ignored](https://github.com/la-ldaf/ldaf-site/pull/196/files?w=1).

src/routes/about/organization/[slug]/+page.server.ts
```diff
@@ -0,0 +1,134 @@
...
+      if (matchedOfficePage) {
+        return matchedOfficePage;
+      }
+    }
+    throw error(404);
+  } else {
+    return {
+      sys: { id: "0", environmentId: "", spaceId: "" },
+      contentfulMetadata: { tags: [] },
+      pageTitle: "Sample Office Page",
```
Comment (LouisFettet) [inline src/routes/about/organization/[slug]/+page.server.ts:79]: Wasn't really sure where to put this test content since it doesn't match the current pattern like `Header` and `Nav` do (those have components, but the Office Page isn't a component). Open to suggestions here!

src/variables.scss
```diff
@@ -60,11 +60,11 @@ $warning: #ffb105;
...
-$success-lighter: #fff4bf;
-$success-light: #ffe065;
-$success: #ffb105;
-$success-dark: #e5a000;
-$success-darker: #936620;
+$success-lighter: #dcfcdc;
+$success-light: #70e17b;
+$success: #00a91c;
+$success-dark: #33803f;
+$success-darker: #3c643b;
```
LouisFettet, line 67: In Figma these were showing the correct colors but the hex code labels were wrong. They've since been updated [here](https://www.figma.com/file/oGKbyCnCRRdNzLYbiags93/LDAF-Component-Library-USWDS-3.0.2?type=design&node-id=2305-12131&t=CRvVxEVyuShcrMuq-4).

src/lib/components/CopyToClipboard/CopyToClipboard.svelte
```diff
@@ -0,0 +1,74 @@
+<script lang="ts">
```
LouisFettet, line 1: I figured it made sense to break this out into a separate component since it could be useful elsewhere. I wasn't sure if it needed its own story though, so I only added tests (it can be seen in the `ContactInfoBox` story though).

LouisFettet (commented):
Self-review!

src/lib/components/ContactInfoBox/ContactInfoBox.svelte
```diff
@@ -0,0 +1,56 @@
...
+    </Section>
+  {/if}
+  {#if validContacts}
+    <strong class="display-block margin-bottom-1">Contact</strong>
+    {#each validContacts as { sys, entityName, phone, phoneExt, email } (sys.id)}
+      <Section>
+        <svelte:fragment slot="label">{entityName}</svelte:fragment>
+        <svelte:fragment slot="info">
+          {#if phone}
+            <Link href="tel:+1{removePhoneFormatting(phone)}{phoneExt ? `;${phoneExt}` : ''}">
```
LouisFettet, line 44: Definitely rabbit-holed on this a bit; from what I read it sounds like using a `;` is the most widely-supported way to tell a phone application to wait before dialing the extension.

src/routes/layout.scss
```diff
@@ -0,0 +1,22 @@
...
+   USWDS Source: 
+    https://github.com/uswds/uswds/blob/c397204e35a20cc31e9a2da9567cac8d0214264a/packages/usa-header/src/styles/_usa-header.scss#L58-L64
+*/
+@include at-media("desktop") {
+  .usa-header {
+    border-bottom: 1px solid $base-lighter;
+    // Override with no border on next sibling.
+    & + .usa-section,
+    & + main {
+      border-top: none !important;
```
Benaiah, line 19: I'd _really_ prefer not to use `!important`, it's a serious maintainability issue. Does placing this CSS at the end of `src/app.scss` override it consistently? It's not perfect since the CSS is a little further away from the code it's affecting but I think it's a lesser evil than `!important`.

src/lib/components/ContactInfoBox/ContactInfoBox.svelte
```diff
@@ -0,0 +1,56 @@
...
+  import "./ContactInfo.scss";
+  import Section from "./ContactInfoSection.svelte";
+  import CopyToClipboard from "$lib/components/CopyToClipboard";
+  import Link from "$lib/components/Link";
+  import classNames from "$lib/util/classNames";
+
+  import type { Maybe, ContentTypeLocation, Contact } from "$lib/services/contentful/schema.js";
+
+  export let address: Maybe<ContentTypeLocation> | undefined;
+  export let contacts: Maybe<Contact>[] | undefined;
```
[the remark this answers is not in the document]
Benaiah: I think either way works, and we'll probably get a better idea of how to do it as we further integrate Contentful. The duplication of blank properties in tests and stories can be handled by using something like `Pick`/`Omit` in the type to narrow down what the type requires, or by just using a helper function that converts a short form of the object into the full object necessary.

Benaiah (changes requested):
The only blocker I see here is the use of `!important` which I would like to avoid. Other than that I only have small comments or things that can be addressed either in this PR or in a followup (like removing some of the duplication in the tests and stories and moving test data out of the page file).

src/lib/components/ContactInfoBox/ContactInfoBox.svelte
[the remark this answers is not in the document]
LouisFettet: Good suggestion! I always forget about how useful `Pick` and `Omit` are.

src/routes/about/organization/[slug]/+page.server.ts
Comment (LouisFettet) [reply src/routes/about/organization/[slug]/+page.server.ts]: I like this suggestion; turns out that SvelteKit will auto-ignore any underscore-prepended directories under `routes/` so I'll opt for the pattern we've been using elsewhere and move the sample response into `__tests__/OfficePageTestContent.ts`

src/routes/layout.scss
```diff
@@ -0,0 +1,22 @@
...
+);
+
+/* For some reason the border between the header and the rest of the page is 
+    placed on the next sibling of the header, but only if it's a .usa-section
+    or <main>, presuambly since those are full-width elements. Since there's no
+    comments or guidance around this we'll override this rule and apply it to
+    the bottom of the header itself, which seems like a more sane approach.
+   USWDS Source: 
+    https://github.com/uswds/uswds/blob/c397204e35a20cc31e9a2da9567cac8d0214264a/packages/usa-header/src/styles/_usa-header.scss#L58-L64
+*/
```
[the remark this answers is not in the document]
LouisFettet: Yeah, I suppose that makes sense. But then again there's the `usa-header--extended` class which probably should have handled this. It is what it is; we've got the border working with this override and it's documented in case things break in the future.

src/lib/components/ContactInfoBox/ContactInfoBox.svelte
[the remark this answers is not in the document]
LouisFettet: Resolved in 5316af2.

src/lib/components/ContactInfoBox/ContactInfoBox.svelte
```diff
@@ -0,0 +1,56 @@
...
+
+  $: validContacts = contacts && contacts.filter((contact): contact is Contact => !!contact);
```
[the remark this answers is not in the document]
LouisFettet: Resolved in 26ef215.

src/routes/layout.scss
LouisFettet: Good call, I hadn't thought to move it into `src/app.scss`, but this seems to work well without the need for `!important`. Resolved in c31c45b.

Benaiah (approved):

PR #152, opened 2023-05-19
Fully typed GraphQL queries

Benaiah:
Depends on #151.

## Proposed changes

- TODO

PR #204, opened 2023-06-03
GraphQL contentful client

Benaiah:
Based on #152.

## Proposed changes

- Adds Contentful client creation that allows configuring the `fetch` method.


src/routes/test-contentful-content/+page.server.ts
```diff
@@ -16,11 +17,22 @@ const query = gql`
...
+    token: CONTENTFUL_DELIVERY_API_TOKEN,
+  });
+  const data = await client.fetch<EntryQuery>(printQuery(query));
   if (data) {
-    return data?.testRichText?.body?.json;
+    const document = data?.testRichText?.body?.json as Document | undefined | null;
+    return {
+      query,
+      document: document || markdownDocument.document,
+    };
```
LouisFettet, line 31: What's the reasoning behind returning `query` here? I'm not seeing it used in the corresponding `+page.svelte`.

src/lib/services/contentful/graphqlClient.ts
```diff
@@ -0,0 +1,63 @@
...
+
+type ClientOptions = {
+  spaceID: SpaceID;
+  token: Token;
+  apiPrefix?: APIPrefix;
+};
+
+const delim = "#" as const;
+
+type ClientKey = `${APIPrefix}${typeof delim}${SpaceID}${typeof delim}${Token}`;
```
LouisFettet, line 15: TIL you can make a template literal type; that's pretty cool.

src/lib/services/contentful/graphqlClient.ts
```diff
@@ -0,0 +1,63 @@
...
+        headers: {
+          "Content-Type": "application/json",
+          Authorization: `Bearer ${token}`,
+        },
+        body: JSON.stringify({ query }),
+      });
+      if (!response.ok) {
+        throw new Error(
+          `Got failed response with status code ${response.status} ${response.statusText}`
+        );
```
LouisFettet, line 52: How do errors like this get handled? Do we need a try / catch rather than an if / else on the response in a route like `[...dynamicRoute]/+page.server.ts`?

src/lib/services/contentful/graphqlClient.test.ts
```diff
@@ -0,0 +1,42 @@
+import { describe, expect, it, vi, type Mock } from "vitest";
+import getClient, { type Client } from "./graphqlClient";
+
+const query = "{ some { graphql { query } } }";
+
+const sampleData = Symbol("sample return value");
```
LouisFettet, line 6: Just curious, why do we use `Symbol` here?

LouisFettet (approved):
I was going to say that I don't love having to use the same boilerplate before making every request, but it looks like you resolve this in #209, so this is good by me as a working midway-point until we merge that PR as well.

I assume the `contentful` package just didn't quite suit our needs here? I figured it might not stick around since it didn't support GraphQL, but I was hoping that it would provide some useful and helpful features for setting up previewing.

Got a few other questions below, but overall this LGTM! Would be good to get @hinzed1127's eyes on this as well though.

src/routes/test-contentful-content/+page.server.ts
Benaiah: This is an artifact of when the previewing was going to use the query to request the preview data on the front-end. I'll remove it.

src/lib/services/contentful/graphqlClient.test.ts
Benaiah: It was so that we could compare object identity while still having a value that's immediately understandable in console logs. We could use `{}` for the same effect, but when logged to the console it wouldn't immediately be obvious what that value represented.

hinzed1127 (approved):

src/routes/test-contentful-content/+page.server.ts
hinzed1127: ^ don't forget about this!

src/lib/services/contentful/graphqlClient.ts
Benaiah: This will cause a 500 currently, but the follow-up PR to add server-side previewing also addresses this.

PR #209, opened 2023-06-05
Server side previewing

Benaiah:
Jira ticket: [LDAF-201](https://ldaf.atlassian.net/browse/LDAF-201)

We want to allow content editors to view the site with unpublished changes applied.

## Proposed changes

- Centralizes Contentful client creation into a single place in a server hook called `handleToken` and uses `locals` to pass it to the server-side `load` functions.
- In the `handleToken` server hook, if the `preview` query parameter is present and an access token is provided (either as a `ldafUserToken` cookie or as an `Authorization: Bearer ...` header), creates a Contentful client that uses the preview API token and automatically adds a `preview` variable with the value `true` to all GraphQL requests.
- Uses GraphQL for all content requests; removes the Contentful REST client API.
- Adds an OAuth login flow with Contentful allowing users to sign in to the site and have their token stored in a cookie to be used to authenticate with Contentful and authorize them to access unpublished content in the given space (see "Possible drawbacks" for some negative aspects of this approach).
- Adds Content-Security-Policy headers that allow Contentful to embed the site (this doesn't work on preview deploys when Vercel deploy protection is turned on).
- Adds a basic login/logout UI to the page header.
- Adds authentication error handling.

<details>

<summary>

## Screenshots

</summary>

Logged out:

image

Upon clicking the "Login" button:

image

After approving the login on Contentful's side and going thru the login flow:

image

Video of the login flow:

https://github.com/la-ldaf/ldaf-site/assets/1425133/8b116c0c-cd81-4a02-be34-1537c35da4e9

</details>

- [x] Manually tested logging in and out and accessing unpublished versions of entries
- [ ] Tests added to cover the change

### Alternate solutions

- We could store the user's management API token and "whether they have been authenticated or not" in a persistent cache like Redis. This would have the advantage of not exposing a Contentful Management API token in a JS-accessible cookie that has to use `SameSite=none` to work in the Contentful preview pane.

- We could make the cookie inaccessible to JS by introducing another request on login whose response includes `Set-Cookie` with the `HttpOnly` flag.

### Possible drawbacks

- Contentful's preview pane embeds the site as an `iframe`, so the cookie returned by the OAuth login flow must be set to `SameSite=None` to include it in cross-origin requests (which includes requests made by an `iframe` embedded in another domain). On top of that, the current implementation uses a JS-accessible cookie. This is rather insecure, because we are sending the authorization cookie on requests to third-party domains, and any JS in the page has access to the API token. To make things worse, this is a Contentful Management API token (not a delivery token), so it has write access to Contentful equivalent to the user it was created for.

## Requested feedback

- This is a security sensitive PR, so close inspections of the security of the PR and suggestions to improve the security of the implementation are very much wanted.

[LDAF-201]: https://ldaf.atlassian.net/browse/LDAF-201?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Benaiah:
Closing in favor of the rebased version (with a lot more added), #274.

PR #212, opened 2023-06-06
Add search page

hinzed1127:
Jira ticket: [LDAF-309](https://ldaf.atlassian.net/browse/LDAF-309)

## Proposed changes

- Adds `/search` subroute
- Adds in algolia search dependencies, including [`InstantSearch.js`](https://www.algolia.com/doc/guides/building-search-ui/what-is-instantsearch/js/), which is Algolia's UI library that uses vanilla JS
- Searches made in the global header search bar route to `/search` with the given query then querying algolia

- [x]  Search page exists and matches preliminary designs
- [x]  Search results are getting served via a connection to Algolia indices

### Alternate solutions

We could roll our own UI around the [Algolia API](https://www.algolia.com/doc/rest-api/search/) and its JavaScript search client, but that seems redundant given how much Algolia provides out of the box in terms of things that sync up within a UI.

### Possible drawbacks

Using `InstantSearch.js` requires some client-side initialization, where a connection to Algolia is made and then all of the UI components are injected into the page. This felt pretty snappy when testing out on a deployed site, but it would be nice to get some additional performance testing in

## Requested feedback

I'm mostly curious how y'all feel about the UI components and how that is generally set up and functions

[LDAF-309]: https://ldaf.atlassian.net/browse/LDAF-309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app.scss
```diff
@@ -4,3 +4,17 @@
...
+}
+
+body {
+  display: flex;
+  flex-direction: column;
+}
+
+main {
+  flex: 1;
+}
```
hinzed1127, line 20: this is to push the copyright tag to the bottom of the page. Probably not much of an issue in the final state of pages, but on stubs or pages with small `<main>` sections, the footer is in the middle of the page

src/routes/search/+page.svelte
```diff
@@ -0,0 +1,111 @@
+<script lang="ts">
+  import "./search-page.scss";
+  import { onMount } from "svelte";
+  import algoliasearch from "algoliasearch";
+  import instantsearch from "instantsearch.js";
+  import { configure, hits, searchBox, stats, pagination } from "instantsearch.js/es/widgets";
+  import { searchHitsTemplate } from "./searchHelpers";
+
+  const ALGOLIA_APP_ID = "9IIDYROXZ5";
+  const ALGOLIA_API_KEY = "0eac885faaa4c70dfc9dd8b6ab4ab10f";
```
hinzed1127, line 10: per the Algolia [docs](https://www.algolia.com/doc/guides/security/api-keys/#search-only-api-key):
> Algolia provides a search-only API key that allows you to search your data. It works on all the indices of your app and is safe to use in your production frontend code. Nevertheless, you may want to use this key to create more restricted API keys.

src/routes/search/+page.svelte
```diff
@@ -0,0 +1,111 @@
...
+  onMount(() => {
```
hinzed1127, line 11: I mentioned this in the PR description, but given that all of the search elements are injected into the page here, it probably makes sense to include a loading spinner/indicator for potentially slower loads

src/routes/search/+page.svelte
```diff
@@ -0,0 +1,111 @@
...
+    const search = instantsearch({
+      searchClient: algoliasearch(ALGOLIA_APP_ID, ALGOLIA_API_KEY),
+      // TODO: change this the "contentful" index when ready (name subject to change)
+      indexName: "media-sample-data",
+      routing: true,
```
hinzed1127, line 16: As long as the url + search query follows the [expected format of algolia](https://www.algolia.com/doc/guides/building-search-ui/going-further/routing-urls/js/#basic-urls), it will automatically populate the search with the correct search query upon initialization

src/routes/search/+page.svelte
```diff
@@ -0,0 +1,111 @@
...
+          (element as HTMLElement).style.display = helper.state.query === "" ? "none" : "";
+        });
+
+        helper.search();
+      },
+    });
+
+    search.addWidgets([
+      configure({
+        hitsPerPage: 3,
```
hinzed1127, line 30: probably will want to increase this in a final state

src/routes/search/searchHelpers.ts
```diff
@@ -0,0 +1,23 @@
+// import type { Hit } from "@algolia/client-search";
+// import type { AlgoliaHit, HitHighlightResult } from "instantsearch.js";
```
hinzed1127, line 2: I'm still having some TypeScript woes on this page trying to define the expected shape of the search hits

hinzed1127 (commented):
Self-review

src/routes/search/searchHelpers.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export function searchHitsTemplate(hit) {
```
hinzed1127, line 4: this will need to be updated to match the shape of the data from Contentful. Right now, it's based on the `media-sample-data` index hits

src/routes/search/+page.svelte
```diff
@@ -0,0 +1,111 @@
...
+        attributesToSnippet: ["content:14"],
```
LouisFettet, line 31: Is there a reason we're marking this lower than the default of `80`?

src/routes/search/+page.svelte
```diff
@@ -0,0 +1,111 @@
...
+          text(data, { html }) {
+            let count = "";
+
+            if (data.hasManyResults) {
+              count += `${data.nbHits} results`;
+            } else if (data.hasOneResult) {
+              count += `1 result`;
+            } else {
+              count += `No result`;
+            }
```
LouisFettet, line 72: nit: I don't think we need to append here (can just replace / reassign), unless there's a TS warning / error?

LouisFettet (commented):
This is looking great! Got a lot of miscellaneous thoughts below. Mostly suggestions for `TODO` comments and follow-up tickets.

src/app.scss
Benaiah: @LouisFettet's suggestion seems like the better choice just in case `main` isn't set up as expected, assuming the `position: sticky` is widely-supported enough or falls back to an acceptable behavior.

src/routes/search/+page.svelte
Benaiah, line 72:
```suggestion
            let count = "No results";
            if (data.hasManyResults) {
              count = `${data.nbHits} results`;
            } else if (data.hasOneResult) {
              count = `1 result`;
            }
```

src/routes/search/+page.svelte
Benaiah: Yeah, I'd love to have more confidence that this won't result in a big bill or a website with degraded functionality if someone runs wild with this. Also, even if these are public, they should be defined as environment variables instead of being hardcoded so that they can be quickly rotated without a full change process.

Benaiah (changes requested):
See #245 for a PR fixing the type errors you ran into!

It would be nice to have a writeup (just for internal use, so doesn't need to be fancy) going over the search strategy listing the improvements we want to make to the search component in the future. Later on it could be expanded to discuss how we get data from Contentful to Algolia. Some of the things I'm thinking of:

1. Autocomplete for the search bar in the header
2. Support for running a search on the server
3. Support for submitting a search without JS (or, more realistically, when some of the JS has failed to load) (Requires point 2)
4. Lazy loading (on search field focus) of library that runs autocomplete in the header search bar (Requires point 1)
5. Potentially proxying all Algolia queries thru a server endpoint so that we can hide the API key

None of these are locality-specific enough to fit in a TODO comment, but they're things we want to aim for.

src/app.scss
hinzed1127: yup, that works! I dig it, making that change.

src/routes/search/+page.svelte
hinzed1127: oh yeah, that's a good point

src/routes/search/+page.svelte
hinzed1127: I think I was just trying to slim down the results as I was working with it at first, but that might just be a product of the sample data. E.g., if we show `hit._highlightResult.content.value` instead of `hit._snippetResult.content.value`, it's a bit more verbose:
2e81d6052505577e272520b354e15cca

hinzed1127: ^ I don't mind switching it back to that for the time being. I imagine the LDAF data will be more concise and this problem might not even exist for it

src/routes/search/+page.svelte
hinzed1127: > Yeah, a loading indicator would be useful. Doesn't necessarily need to happen in this PR if you want to write a follow-up ticket for it.

I grabbed one from [here](https://projects.lukehaas.me/css-loaders/) as a quick addition. Worth running by @getpunched but should be fine as a placeholder, at least

src/routes/search/+page.svelte
```diff
@@ -0,0 +1,111 @@
...
+          link: "usa-pagination__button", // the link elements.
+        },
+        templates: {
+          next: "Next",
+          previous: "Previous",
+        },
+      }),
+    ]);
+
+    search.start();
```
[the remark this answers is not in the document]
hinzed1127: so that method name is a little misleading and is just initializing the whole `instantsearch` apparatus and not necessarily making a query. The search function that wraps around queries is [here](https://github.com/la-ldaf/ldaf-site/pull/212/files#diff-4fd57f831561a3f8ff467da7008ad85aee43203174f0bde4e9a5c9227f9f2153R18-R31) and won't make a request if there's no query. The default just gives you whatever's first: 
> If no query parameter is set, the textual search will match with all the objects.

src/routes/search/+page.svelte
```diff
@@ -0,0 +1,111 @@
...
+        hitsPerPage: 3,
+        attributesToSnippet: ["content:14"],
+        snippetEllipsisText: " [...]",
+      }),
+      hits({
+        container: ".hits",
+        templates: {
+          empty: "No results found.",
+          item(hit) {
+            return searchHitsTemplate(hit);
```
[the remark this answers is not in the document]
hinzed1127: Yeah I think that's worth examining in some followup work. I agree it would be nice to utilize Svelte syntax directly

hinzed1127:
@Benaiah The one thing I still need to resolve in the scope of this PR is a few more TypeScript errors, but re search strategy moving forward, here's a [quick writeup ](https://gist.github.com/hinzed1127/c29127d65969a8e6ab649a309c09e968) covering the points you raised.

src/routes/search/search-page.scss
```diff
@@ -0,0 +1,44 @@
+@media (min-width: 40em) {
```
LouisFettet, line 1: Any chance we can use one of our existing breakpoints instead? Same goes for the other media query below.

src/routes/search/search-page.scss
```diff
@@ -0,0 +1,44 @@
...
+
+@media (min-width: 30em) {
+  .search-results .usa-input {
+    border-radius: 5px 0 0 5px;
+  }
+  .search-results .usa-search__submit-icon {
+    display: inline-block;
+    fill: white;
+  }
+}
```
LouisFettet, line 34: Not sure if this block is working as expected. On mobile the icon turns grey, and I feel like the border radius should be consistent on all screen sizes.

src/routes/search/+page.svelte
LouisFettet: This might go a bit too far in the other direction; I think we should try to limit the highlighted / snippeted content to just a few lines, but we can handle that once we're working with real data.

LouisFettet (approved):
LGTM, I think this is a good start and is worth merging now-ish and improving upon later. It would be good to get @Benaiah's approval here as well.

Only thing I think needs to change before merge are a couple things I pointed out w/r/t styling below.

src/routes/search/+page.svelte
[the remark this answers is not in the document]
Benaiah: I think we should definitely consider debouncing in the future if we see a need for it, but the default Algolia behavior should be fine for now.

Benaiah (approved):
LGTM once Louis' styling concerns are addressed! Approving so I don't block this any further.

src/routes/search/+page.svelte
hinzed1127: yeah, he's down with a [smaller color update](https://adhoc.slack.com/archives/C04EL8QK5FU/p1689362806331569). Those are applied [here](https://github.com/la-ldaf/ldaf-site/pull/212/commits/1202edd0eceb5d6b2192ffe046f43d0eb2328f47).

src/routes/search/search-page.scss
hinzed1127: Yeah I can't remember what the motivation for the media query was, but I don't think it has a purpose. Keeping this styles but removing the query.

PR #219, opened 2023-06-08
Run prebuild before dev

Benaiah:
## Proposed changes

Runs `npm run prebuild` before `npm run dev` starts. This prevents errors due to missing JSON "test" documents that we actually rely on in development when the Contentful connection isn't present.


LouisFettet (approved):
LGTM!

PR #233, opened 2023-06-14
Remove Chromatic in favor of publishing Storybook to Vercel.

LouisFettet:
Pretty straightforward; we're no longer using Chromatic due to rate limits, so we might as well remove the CLI (and stop getting dependabot updates for it). Afterwards I'll remove the API token from our secrets storage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,12 +31,6 @@ To run Storybook locally:
 npm run storybook
 ```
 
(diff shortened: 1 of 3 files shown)


Benaiah (approved):

LouisFettet:
Environment variable removed from Vercel!

PR #245, opened 2023-06-21
Fix type errors in Dan's Algolia PR

Benaiah:
## Proposed changes

- Updates instantsearch.js to get rid of spurious type error
- Properly types search template


Benaiah:
Having trouble resolving the unit test errors in this but I've got the types working.

PR #266, opened 2023-06-27
Banner components

hinzed1127:
Jira ticket: [LDAF-292](https://ldaf.atlassian.net/browse/LDAF-292)

## Proposed changes

- Adds an `AnnouncementBanner` and a `DotGovBanner` component
- `AnnouncementBanner` just links to another page, whereas `DotGovBanner` is the existing top banner, just replacing the US flag with the LDAF logo and rephrasing federal gov -> LA gov

**Announcement Banner**
b949412699ee4331a1704fd05dd95aa2

**DotGov Banner**
0632e5c4a9ee634e1f90f2cc7004b20f

- [x]  Fulfilled ACs
- [x] Tests included

## Other details

It's worth noting that the `DotGovBanner` component might not even get utilized in the final site design. Since it's already the only design for the USWDS `Banner` component, it's easy to throw in with the Louisiana copy edits.

The Announcement banner is our modified (and much visually simpler) version that should be what actually gets used from time to time.

### Alternate solutions

I broke these out into separate components, mainly because:
1. I don't anticipate the `DotGovBanner` really changing. It's really just a static drop-in component
2. `DotGovBanner` is technically an accordion, whereas is `AnnouncementBanner` is much simpler, with no needs for a toggle state. There's not really a lot of a overlap between the 2, so preserving it as a single component felt like unnecessary complexity

---

I broke out `AnnouncementBanner` into 3 props:
 1. `headerText`
 2. `linkText`
 3. `href`
 4. `ariaLabel`

This could also be made a `slot`, but I don't think the rigid design is a negative in here, since its flexibility is probably intentionally limited.

## Requested feedback

[LDAF-291]: https://ldaf.atlassian.net/browse/LDAF-291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[LDAF-292]: https://ldaf.atlassian.net/browse/LDAF-292?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/lib/components/AnnouncementBanner/AnnouncementBanner.scss
```diff
@@ -0,0 +1,35 @@
+.usa-banner.ldaf-banner--announcement {
+  background: $primary-darkest;
+
+  .usa-banner__header-text,
+  .usa-banner__guidance {
+    color: white;
```
LouisFettet, line 6: nit: I know this isn't necessary, but we should probably use a variable here for consistency's sake.

src/lib/components/AnnouncementBanner/AnnouncementBanner.scss
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  .usa-banner__button-text {
+    color: $primary-lighter;
+  }
+
+  .ldaf-logo {
+    filter: invert(1);
+  }
```
LouisFettet, line 15: What is this doing? It doesn't seem to be inverting the logo colors for me, if that's the intention.

src/lib/components/AnnouncementBanner/AnnouncementBanner.scss
```diff
@@ -0,0 +1,35 @@
...
+    padding-left: 0.25rem;
+    color: $primary-lighter;
+    display: flex;
+    align-items: center;
+  }
+
+  .ldaf-banner--announcement-text {
+    display: flex;
+    align-items: center;
+  }
```
LouisFettet, line 27: Not sure if this is the appropriate block, but it looks like we're not properly centering the announcement as shown in the Figma mock; both the DotGovBanner and AnnouncementBanner look like they're left-aligned within the main container.

src/lib/components/AnnouncementBanner/AnnouncementBanner.svelte
```diff
@@ -0,0 +1,33 @@
...
+  import { url as arrowIcon } from "$icons/arrow_forward";
+  import classNames from "$lib/util/classNames";
+
+  export let headerText = "Here's a site-wide announcement!";
+  export let linkText = "Read more here";
+  export let href = "/";
+  let classes = classNames("usa-banner", `ldaf-banner--announcement`);
+</script>
+
+<section id="top" class={classes} aria-label="Official website of the Louisiana state government">
```
LouisFettet, line 15: Probably need to change this `aria-label` so that it's something generic, and maybe make this something we can optionally set as a prop.

src/lib/components/AnnouncementBanner/AnnouncementBanner.svelte
```diff
@@ -0,0 +1,33 @@
...
+  <header class="usa-banner__header">
+    <div class="usa-banner__inner">
+      <div class="grid-col-auto">
+        <img
+          class="usa-banner__header-flag ldaf-logo"
```
LouisFettet, line 20: The `ldaf-logo` class probably isn't appropriate here, right?

LouisFettet (commented):
Got a few minor comments, mostly around styling. Overall looks good though! Did we confirm with @getpunched that the typical announcement banners won't ever need collapse/expand functionality? If so I think it makes perfect sense to break these components out as you've done here.

src/lib/components/AnnouncementBanner/AnnouncementBanner.svelte
hinzed1127: yeah, that was an accidental carryover from splitting the 2 banners up. Renaming that to be more accurate.

src/lib/components/AnnouncementBanner/AnnouncementBanner.scss
hinzed1127: this is another case of the Icon component not supporting recoloring (see [LDAF-327](https://ldaf.atlassian.net/browse/LDAF-327)). Since this is just going from black to white, it's simpler than the `filter` we needed to use for the copy button. I'll add a TODO, as well

src/lib/components/AnnouncementBanner/AnnouncementBanner.scss
hinzed1127: Courtney made a few adjustments to the design which are reflected here now, as well. Now, it's as follows:
- `DotGovBanner` is always left-aligned
- `AnnouncementBanner` is left-aligned on mobile, centered on on anything wider

src/lib/components/AnnouncementBanner/AnnouncementBanner.scss
LouisFettet: Ah I gotcha. Thanks for adding those notes!

src/lib/components/AnnouncementBanner/AnnouncementBanner.scss
LouisFettet: Seems like the `AnnouncementBanner` is still centered on mobile (at least on the `/news` page). I think it looks fine but based on your comment you might still need to make an update to get it left-aligned on mobile.

LouisFettet (approved):
LGTM, just need to make an update if we want to left-align `AnnouncementBanner` on mobile.

src/lib/components/AnnouncementBanner/AnnouncementBanner.scss
hinzed1127: good catch, should be fixed now

PR #273, opened 2023-06-30
LDAF-350 Set up foundation for future navigation and routing.

LouisFettet:

This is an initial PR to set us up with using the Page Metadata content type in Contentful as the source of truth for all future navigation and routing logic.

## Proposed changes

- Adds support for supplying a Contentful environment when creating a client.
- Adds a `pageMetadataMap` for use across all pages and layouts, which can be used to fetch data about a page, including the URL, parent and child pages, the page's title (for use when linking to the page), etc.
- Adds stubbed routes for Top Tier pages, Service Group pages, an Office aggregator page, a Boards and Commissions aggregator page, and individual Board/Commission pages.
- Makes some related adjustments to the route for Office pages.
- Removes our catch-all `/[...dynamicRoute]` as the service group route now takes precedence.

I opted to skip stubbing out routes for News, Events, Licensing, and Laws since those are still a bit in flux. Most of these will operate as aggregators with individual record pages under them, similar to how `/about/organization` and `/about/boards-commissions` are set up here.

This PR roughly accomplishes one of the A/C from the ticket:
* [x] Stubbed page for every page metadata entry

## Requested feedback

See self-review.


src/routes/[topTierPage]/+page.server.ts
```diff
@@ -0,0 +1,48 @@
...
+            slug
+          }
+        }
+      }
+    }
+  }
+`;
+
+export const load = async ({ parent, params }) => {
+  const { pageMetadataMap } = await parent();
```
Comment (LouisFettet) [inline src/routes/[topTierPage]/+page.server.ts:27]: The issue with this pattern is that we're effectively calling `loadPageMetadataMap` every time we load (or preload) a page, which is pretty terrible. I'd be happier with a solution where this only gets called once per session on the first page load, but I think ideally we could store this map somewhere for use by all sessions and only update it when Contentful gets updated. 

What do y'all think is the best way to make a small improvement here? I'm thinking about trying to use either `store` or `setContext` / `getContext` but I'm not sure which one would be a better fit here.

src/routes/about/organization/[officePage]/+page.server.ts
```diff
@@ -47,37 +45,47 @@ const query = gql`
...
+          }
+        }
       }
     }
   } else {
-    return OfficePageTestContent;
+    return { officePage: OfficePageTestContent, pageMetadata: {} };
   }
   throw error(404);
-}) satisfies PageServerLoad;
```
Comment (LouisFettet) [inline src/routes/about/organization/[officePage]/+page.server.ts:83]: I removed this since it didn't seem necessary, but I can re-add it (along with adding `PageServerLoad` to other `+page.server.ts` files and `LayoutServerLoad` to `+layout.server.ts` files) if y'all think we should have it.

src/routes/[topTierPage]/+page.server.ts
Comment (hinzed1127) [reply src/routes/[topTierPage]/+page.server.ts:27]: I think that the Context API is the way to go here, primarily because:

1. `pageMetadataMap` is not reactive, so we don't need to subscribe to it like a store value.
2. since `pageMetadataMap` is initialized from the root, it should be accessible to all other pages via context, since they'll all be children.

Then we could just store this in either localStorage, or would this perhaps be an appropriate use case for service workers? I haven't worked with them before, so I'm wary of a cache not properly updating and leading to stale data, especially with something like route generation, but in either case, _some_ sort of persistence mechanism should work in tandem with `setContext`/`getContext`

hinzed1127 (approved):
Overall I think this looks good! While I agree that we should cache the request made in`loadPageMetadataMap` ~and load it via Context~ and then load the cached value in child routes, I also don't want that to hold other things up. My 2 cents is to get this merged if the context stuff doesn't come together quickly, and in a follow-up PR, do the follow:
- add caching for the graphQL query to load all the page metadata
- store the metadata map constructed from that response in a cookie (I think this is the right approach, since this is all happening server-side?)

src/routes/[topTierPage]/+page.server.ts
Comment (hinzed1127) [reply src/routes/[topTierPage]/+page.server.ts:27]: Actually...since SvelteKit already passes propagates data from parent `layout`s down to their children, I don't think we even need to worry about using Context. Simply caching the call should be enough

Benaiah (approved):
I did a bunch of work this week looking into how we can avoid constructing the sitemap to avoid performance and stale-cache issues but it hasn't fully panned out yet so I think we should move forward with this in the meantime. Apologies for the long delay on this approval because of that!

src/routes/about/boards-commissions/+page.server.ts
```diff
@@ -0,0 +1,9 @@
+export const load = async ({ parent }) => {
+  const { pageMetadataMap } = await parent();
+  const matchedPageMetadata = [...pageMetadataMap].find(([_, page]) => {
+    return page.url === "/about/boards-commissions";
+  });
+  if (matchedPageMetadata && matchedPageMetadata.length === 2) {
```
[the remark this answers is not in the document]
LouisFettet: Yup, most of what I had originally written was about three times smaller before resolving all the TS warnings 😅

src/routes/loadPageMetadataMap.ts
```diff
@@ -0,0 +1,107 @@
...
+import getContentfulClient from "$lib/services/contentful";
+
+import type { PageMetadataCollectionQuery } from "./$queries.generated";
+
+// extend the type of the items we get back from the query so we can add children and a full URL
+export type PageMetadataMapItem = NonNullable<
+  PageMetadataCollectionQuery["pageMetadataCollection"]
+>["items"][number] & { children?: string[]; url?: string | null };
+
+const pageMetadataMap = new Map<string, PageMetadataMapItem>();
```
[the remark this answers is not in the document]
LouisFettet: I believe that using a `Map` here is just [more performant](https://www.zhenghao.io/posts/object-vs-map), but I'm not opposed to switching to an `Object` if we think that would be better (probably in a follow-up).

LouisFettet:
Appreciae the feedback here y'all! We can talk more tomorrow about how to improve the performance, but for now I'm going to move this forward and merge.

PR #274, opened 2023-07-01
Server-side user authentication+previewing, and server- and client-side error logging

Benaiah:
Jira ticket: [201](https://ldaf.atlassian.net/browse/LDAF-201)

## Proposed changes

- Adds a `handleToken` hook that authenticates provided user tokens (in either the `Authorization` header or an `HttpOnly` `ldafUserToken` cookie) against a Redis instance (set up in production and runnable locally using either `redis-server` with no arguments or `docker compose up`) containing user objects that list the data required to render the current user (which is also sent to the root layout and the browser) and the user's Contentful Management API token (which is not). (Note that the `ldafUserToken` is _not_ the Contentful Management API token.)

- The `handleToken` hook initializes the Contentful client that requests normal Contentful content.

- In the `handleToken` hook, if the request is for a url containing a `preview`, gets the management API token from Redis and makes a call to the Contentful space to ensure the user is still allowed to view preview content in that space. If the authentication+authorization succeeds, we replace the Contentful client with one that uses the preview API token and passes a `preview: true` variable to every GraphQL request (I haven't found a good way to automatically pass the variable as an argument to the queries, so the queries do need to be written with the `preview` variable in mind for previewing to work). 

- Adds `/login` route that has a default form action taking a `POST` request containing a Contentful Management API token, authenticates with Contentful using the token (it only requests the current user's information - the actual authorization to see unpublished content in the space is done on every preview request by the `handleToken` hook), and then stores the user info in Redis to use for further authentication.

- Adds `/logout` route that clears the `ldafUserToken` cookie and (attempts to) delete the stored user info in Redis (this is given an expiration time of one week in Redis itself, so unsuccessful deletes will not leave data around indefinitely).

- Adds an OAuth2 callback at `/callback/oauth2` that receives a request from Contentful containing a Management API token (which is passed in the hash, so the OAuth2 callback requires client-side JS to work - you can still log in without JS by entering a personal access token on the `/login` page) and then requests `/login` from the client side to save the token to Redis. This is the _only_ time that client-side code has access to the Management API token (and the Management API token is not stored in the cookie either, so we're not sending it to third parties in HTTP requests - the cookie has to be `SameSite=None` in order to work inside the iframe used by the Contentful preview in the Contentful interface.)

- Adds a "Login with Contentful" link to the `/login` page that takes you to Contentful, where you can log in and authorize the OAuth app to use your account and then be redirected back to the OAuth2 callback page to save the Contentful Management API token to Redis.

- Adds a server-side and client-side logger that currently logs to the console (in dev or tests) or to the #ldaf-eng-logs Slack channel (in production, including `npm run build && npm run preview`).

- Adds `handleError` hooks to both the server and the client that use the respective logger.

- Sets the appropriate `Content-Security-Policy` to allow embedding the site in the preview iframe in the Contentful interface.

- Adds _lots_ of tests for everything above (except the logger, which was verified manually - it's possible to test the logger as well and we should probably do it eventually but this PR is already quite large).

- Adds a `Dockerfile` and `docker-compose.yml` that allow spinning up other services in addition to the main web app. This can be completely ignored if you don't want to deal with Docker - the site works (or at least should work) exactly the same without Redis running if you're not requesting preview content. Even if you need to work with Redis, you can run `redis-server` with no arguments and then start the app and everything including user sessions and authentication will work.

- [x] Tested manually
- [ ] Full security writeup finished (in progress).
- [ ] Additions made to `CONTRIBUTING.md` explaining how to work with all the new features like logins and logging in development.

## Requested feedback

See my upcoming self-review.


src/hooks.server.ts
```diff
@@ -1,11 +1,252 @@
...
+import getErrorStatus from "$lib/util/getErrorStatus";
+import consoleErrorIfYouCan from "$lib/util/consoleErrorIfYouCan";
 
-export const handle = (async ({ event, resolve }) => {
-  const response = await resolve(event, {
+export const handleError = (({ error, event }) => {
+  const logger = event.locals.logger ?? newLogger();
+  const message = getErrorMessage(error);
+  const status = getErrorStatus(error);
+  (async () => {
```
Benaiah, line 22: This needs to be awaited.

src/hooks.client.ts
```diff
@@ -0,0 +1,28 @@
...
+      try {
+        logger.setPublicContext("url", event.url.toString());
+      } catch (_) {
+        // do nothing
+      }
+      await logger.logError(error);
+    } catch (err) {
+      consoleErrorIfYouCan(`Error while trying to log unexpected error: ${message}`);
+    }
+  };
```
Benaiah, line 22: This function isn't being called?

src/lib/logger/public.ts
```diff
@@ -0,0 +1,84 @@
...
+} from "./types";
+import { logMessage, logError, logErrorResponse } from "./logMethods";
+import { newPublicMessage } from "./logMessages";
+export type { PublicLogger, PublicContext, SetPublicContextOptions };
+
+// Regular setContext _only exists on the server_. On the client, _always_ use
+// setPublicContext. setPublicContext can also be used on the server to set context that will be
+// accessible by the client logger (so it must be serializable!). Note that "public" means "public
+// to the browser session of the current user", so things like the current user data are safe to store
+// in public context because only that user will see them.
```
LouisFettet, line 17: Thanks for this note! Maybe to avoid confusion we could replace `Public` with `UserSession` instead? Not sure if that's any more or less accurate but I am worried that "public" could be misinterpreted when reading it outside of this file.

src/lib/util/catchMeIfYouCan.ts
LouisFettet, line 1: Tom Hanks in Catch Me If You Can

src/routes/+layout.svelte
```diff
@@ -29,20 +40,66 @@
...
+    <main id="main-content">
+      <slot />
+    </main>
+    <!-- TODO: reinstate <Footer /> once designed and implemented -->
+    <footer class="usa-footer padding-top-7 padding-bottom-1 text-center">
+      ©2023 Louisiana Department of Agriculture and Forestry. All rights reserved
+    </footer>
+    <!-- TODO: reinstate <Identifier /> once designed and implemented -->
+    <!-- (or remove if encapsulated by footer) -->
+  {/if}
```
LouisFettet, line 102: nit: we should probably still include the footer when there's a `previewAuthenticationError` right?

src/routes/about/organization/[slug]/+page.server.ts
```diff
@@ -1,18 +1,14 @@
...
 import type { PageServerLoad } from "./$types";
-import type { OfficePage } from "$lib/services/contentful/schema";
-import type { OfficePageQuery } from "./$queries.generated";
+import type { OfficePageQuery, OfficePageQueryVariables } from "./$queries.generated";
 
 const query = gql`
-  query OfficePage {
-    officePageCollection {
+  query OfficePage($slug: String, $preview: Boolean = false) {
+    officePageCollection(preview: $preview, where: { metadata: { slug: $slug } }, limit: 1) {
```
Comment (LouisFettet) [inline src/routes/about/organization/[slug]/+page.server.ts:11]: This will need to be updated to `pageMetadata` after changes in #285.

LouisFettet (commented):
Local issues I'm having aside, this is looking good. Appreciate all the tests and comments! I definitely owe this a more thorough review when I get back from PTO if y'all haven't merged it by then.

I wasn't able to get Docker working, but I was able to get things running with a normal install of Redis. The initial auth is the only thing that's working for me though; I'm able to provide a token and it gives me a success message, but then when redirecting back to a URL with `?preview` I get a 401 error telling me to log in again (this isn't an issue on non-preview pages; everything looks right and I get the welcome message and logout link in the header). Not sure if this is intentional or not for this stage, or if you're seeing something different on your end, so I figured I would call it out.

Also, is there a timeout on the Redis connection? After a few minutes of inactivity I notice that I get a `Socket closed unexpectedly` error with the Redis client. Could be something wrong with my install if you're not seeing this though.

I've got some other comments below; only other note is that it might be good to update the README with instructions for running with Docker / installing Redis.

src/lib/components/ContentfulLoginLink/ContentfulLoginLink.svelte
```diff
@@ -0,0 +1,44 @@
...
+  type $$Props = HTMLAnchorAttributes;
+
+  const baseURL = PUBLIC_CONTENTFUL_OAUTH_ENDPOINT;
+
+  let loginParams: Record<string, string> = {},
+    loginLink = "";
+
+  if (browser) {
+    // we're using page.subscribe instead of just $page because of a weird error that only happens
+    // in tests
```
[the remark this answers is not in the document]
Benaiah: I tried to replicate the error again to add details and it went away 🤷.

src/lib/logger/public.ts
Benaiah: Hmm, I don't like that name because the backend logging is _also_ for a specific user session. What about replacing `Public` with `Client`?

LouisFettet: I like `Client`!

src/routes/[...dynamicRoute]/+page.server.ts
```diff
@@ -23,18 +20,19 @@ const query = gql`
   }
 `;
 
-export const load = (async ({ params }): Promise<DraftNavigationLink> => {
+export const load = (async ({
+  params,
+  locals: { contentfulClient },
+}): Promise<{ text: string }> => {
+  // TODO: create fallback fixture
```
Comment (hinzed1127) [inline src/routes/[...dynamicRoute]/+page.server.ts:27]: just highlighting this in case we want to specifically capture this in a follow-up story

hinzed1127 (commented):
Thanks for all your work on this, it's a lot but looks pretty solid. I had a few small documentation things to note, but otherwise not a whole lot jumped out.

Local development is broken right now due to some of the nav changes (I'm currently getting a 500 page with "got malformed submenu"), which I believe is due to some of the recent navigation updates. I'll give things another go locally to make sure things run as expected once that's sorted out, but other than that, stuff looks good 👍

PR #278, opened 2023-07-04
Basic header image optimization

Benaiah:
## Proposed changes

- Optimizes the logo by resizing it, running it thru pngcrush, then setting up a `srcset` that only loads the high-res version on 2x-DPI screens.
- Optimizes the header background by converting it to JPEG, running it thru jpegoptim, then using the `Image` component to lazy-load it with a blurred background.
- Adds avif and webp support to all the static images I could find.

- [x] Manually tested


src/lib/components/Header/Header.svelte
```diff
@@ -42,7 +60,7 @@
 
       <a class="ldaf-logo__compact" href="/">
         <!-- TODO: Replace alt value with content from CMS. -->
-        <img src={ldafLogo} alt="Louisiana Department of Agriculture and Forestry Home" />
+        <Logo />
```
LouisFettet, line 63: nit: we can remove the `TODO` comment since you moved it into the `Logo` component.

LouisFettet (approved):
Just have a minor note that it seems like there's a bit more shufffling that's happening in the header on initial load; the logo doesn't seem like it's being "placeheld" (is that a word?) so there's some shifting that happens when it loads in.

Not a big deal though since it only impacts desktop and loads in pretty quickly. If you think the optimizations here are worth this flash, then feel free to merge!

Benaiah:
Great catch on the layout shuffling, that's fixed now! I'll need to double check my contentful asset rendering PR to make sure that's working correctly once this is merged, but we shouldn't be having any layout shifting anymore.

PR #279, opened 2023-07-04
Render contentful references and assets

Benaiah:
Jira ticket: [LDAF-248](https://ldaf.atlassian.net/browse/LDAF-248)

## Proposed changes

- Renders Contentful rich text images with the Image component
- Adds an `/api/v1/blurhash` route that creates blurhashes on-demand from Contentful images and caches them in Redis

## Screenshots

image

- [x] Manually tested

### Alternate solutions

We considered using Imgix for blurhashes, and we could still easily switch to that, but the API endpoint included here was a small lift and means we don't have to worry about the number of images we have or go thru procurement for anything.


src/hooks.server.ts
```diff
@@ -1,6 +1,48 @@
...
+  // we intentionally don't await this promise here, so that other things can happen while redis is
+  // initializing. Anything that needs redis can await the promise by calling
+  // event.locals.getConnectedRedisClient() and awaiting the result. If nothing awaits the promise
+  // then we never wait on redis connecting first.
+  const redisClientPromise = !KV_URL
+    ? Promise.reject(new Error("could not get redis client"))
+    : createConnectedRedisClient({ url: KV_URL, useTLS: !KV_URL.startsWith("redis://localhost") });
+
+  // So we don't get unhandled promise rejection warnings if nothing awaits the promise
+  redisClientPromise.catch((err) => console.error(err));
```
Benaiah, line 38: If #274 is merged first, this will need to be updated to use the `logger` added in that PR.

src/routes/test-contentful-content/+page.server.ts
```diff
@@ -7,24 +7,60 @@ import type { Document } from "@contentful/rich-text-types";
...
   const data = await client.fetch<EntryQuery>(printQuery(query));
+  const blurhashes = Object.fromEntries(
+    await Promise.all(
+      data?.testRichText?.body?.links.assets.block
+        .filter((item) => !!item)
+        .flatMap(async (item) => {
+          if (!item?.sys?.id || !item?.url || !item?.contentType?.startsWith("image/")) return [];
+          const blurhashResponse = await fetch(`/api/v1/blurhash/${encodeURIComponent(item.url)}`);
+          if (!blurhashResponse.ok) return [];
+          return [item.sys.id, await blurhashResponse.text()];
```
Benaiah, line 56:
```suggestion
          return [[item.sys.id, await blurhashResponse.text()]];
```

src/lib/util/getErrorMessageFromResponse.ts
```diff
@@ -0,0 +1,58 @@
...
+  if (!contentType) return;
+  if (contentType.startsWith("text/plain")) return response.text();
+  if (contentType.startsWith("application/json") || contentType.startsWith("text/json")) {
+    const body = await response.json();
+
+    const bodyMessage = getBodyMessage(body);
+    if (bodyMessage) return bodyMessage;
+
+    const errorMessage = getErrorMessage(body);
+    if (errorMessage) return errorMessage;
```
Benaiah, line 52: `getErrorMessage` always returns a value so `getDataMessage` is not being used right now.

src/lib/util/getErrorMessageFromResponse.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    const dataMessage = getDataMessage(body, { parseDevalue });
+    if (dataMessage) return dataMessage;
+  }
+  return undefined;
```
hinzed1127, line 57: I would add some comments to clarify where the errors are coming from and/or renaming the functions to be more explicit about that. 

Also, since `getDataMessage` always returns a value, there's no need to `return undefined`, so I'd say either remove that or remove the fallback message from `getDataMessage` and put the `return "[could not determine error message!]";` as the return value at the end of this function
```suggestion
    // Check for error messages in order of precedence. Return the first
    // error message that's successfully parsed 

    // Error message from Contentful
    const bodyMessage = getBodyMessage(body);
    if (bodyMessage) return bodyMessage;

    // Error message from SvelteKit
    const dataMessage = getDataMessage(body, { parseDevalue });
    if (dataMessage) return dataMessage;
  }

    // Any other errors. Returns a generic fallback message if
    // it doesn't successfully find an error message  
    const errorMessage = getErrorMessage(body);
    if (errorMessage) return errorMessage;
```

hinzed1127 (approved):
Looks good! I just have one small update around error messaging, but otherwise, LGTM

src/routes/test-contentful-content/+page.server.ts
```diff
@@ -7,24 +7,60 @@ import type { Document } from "@contentful/rich-text-types";
...
+        }) ?? []
+    )
+  );
```
LouisFettet, line 59: Is this something we'll need to do in all of our `+page.server.ts` files, or can this be integrated into the Contentful client?

LouisFettet (approved):
LGTM as well, just a few small notes and questions. Maybe just give us a heads-up after getting the latest updates from `main` so that we can take one more quick look before merging this.

src/routes/test-contentful-content/+page.server.ts
Benaiah: Hmm. The issue with trying to make the Contentful client do it transparently is that it would need to either walk the entire response to every GraphQL request to check for images or rich text, or we would have to tell it about the structure of the query somehow. Since we're parsing the queries at build time and using them as data instead of as strings, it _might_ be possible to do this automatically. Even if that isn't possible, this should get refactored into a helper at the very least.

What I _wish_ we could do is request it in the image component itself, and then have SvelteKit await that on the server before responding to the request. Unfortunately, async server-side rendering is not (yet?) available in SvelteKit.

I'll refactor this into some helper functions to get this merged, and look into doing it transparently in the Contentful client in future work.

LouisFettet:
@Benaiah the `/test-contentful-content` page looks good but unfortunately it looks like you'll need to make some updates to the Core Content pages in order for them to show up. Currently I'm getting a `500` on `/animals/branding` with error `no context was provided for asset hyperlink`

LouisFettet:
Giving this a look now; things seemed rather slow on first run-through but then I changed some `redis-cli` settings to increase memory and buffer limits and everything seemed much smoother (although it's possible that this performance increase was due to having already requested the images on prior run).  

I'm still seeing the `SocketClosedUnexpectedlyError` after about 5 minutes, so I'm going to continue looking into that.

I think the only blocker might be that the image on this page seems to run into an issue where it fails to load, although I haven't looked into why yet: https://ldaf-git-benaiah-render-contentful-references-and-assets-ldaf.vercel.app/plants/pests

src/lib/services/server/kv/index.ts
```diff
@@ -0,0 +1,47 @@
...
+  url = "redis://localhost",
+}: { url?: string } = {}): Promise<Client> => {
+  const parsedURL = new URL(url);
+  // "localhost" is the hostname used when running a redis server locally on the dev machine. "kv"
+  // is the hostname used when running in Docker
+  const useTLS = !(parsedURL.hostname === "localhost" || parsedURL.hostname === "kv");
+  let redisClientError: undefined | unknown;
+  const redisClient = createRedisClient<None, None, None>({
+    url,
+    socket: { tls: useTLS },
```
LouisFettet, line 24: I _think_ we may be able to fix the socket connection error with the `pingInterval` config, although I'm not sure if this could have side effects outside of local development (so we may only want to add this config for local dev). After making this change locally, I didn't see the error pop up after running `dev` for over 10 minutes (usually I saw the error pop up around the 5 minute mark).

```suggestion
    socket: { tls: useTLS },
    // ping every 5 minutes to avoid a socket connection error in local development
    pingInterval: 5 * 60 * 1000,
  });
```

LouisFettet (commented):
Got a question and a minor comment, but otherwise this LGTM! Only blocker is figuring out why we're getting an error on the image on the [/plants/pests](https://ldaf-git-benaiah-render-contentful-references-and-assets-ldaf.vercel.app/plants/pests) page.

src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts
```diff
@@ -101,64 +131,122 @@ const query = gql`
...
+    };
+  };
+  childServiceEntries: (ChildServiceEntry & {
+    description?: ChildServiceEntry["description"] & {
+      blurhashes?: Record<string, string> | null | undefined;
+    };
+  })[];
+  childServiceGroups: (ChildServiceGroup & { url?: string | null | undefined })[];
+  pageMetadata?: ServiceGroupMetadata;
+};
```
Comment (hinzed1127) [inline src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts:168]: Would it make sense to export these from a `src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/types.ts` file and import them here?

src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts
```diff
@@ -45,8 +43,18 @@ const query = gql`
...
+                ...ImageProps
+              }
+              hyperlink {
+                ...ImageProps
+              }
+            }
+          }
         }
-        serviceEntriesCollection(limit: 10) {
+        serviceEntriesCollection(limit: 5) {
```
Comment (hinzed1127) [inline src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts:57]: You might have been hitting query complexity limits, but this'll be too small for some service entries. E.g. `/animals/meat-poultry` currently has 8 service entries, so we'll probably want at that. Ideally figure out with UX what the ceiling on number of service entries might be (10? higher?)

hinzed1127 (approved):
A couple of small comments, but this looks ready to go

src/hooks.server.ts
```diff
@@ -1,6 +1,21 @@
 import type { Handle } from "@sveltejs/kit";
+import { KV_URL } from "$env/static/private";
+import { sequence } from "@sveltejs/kit/hooks";
+import { createClient } from "$lib/services/server/kv";
 
-export const handle = (async ({ event, resolve }) => {
+const handleSetupRedisClient = (async ({ event, resolve }) => {
```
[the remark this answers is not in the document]
Benaiah: When running on Vercel (as opposed to local development) there's no real distinction between "when the app starts" and "when it handles a request" because we're using serverless functions, so the app starts and runs to completion for each request. We can't persist a client between different invocations of a serverless function.

LouisFettet (approved):
LGTM after a quick click-test. 

I'm not seeing blurhashes when navigating between `Core Content` pages that both have hero images (the previous hero stays in place until the new one loads, instead of showing the blurhash first), but we can handle that in a follow-up.

PR #293, opened 2023-07-11
LDAF-350 Create and populate breadcrumbs on all pages using a shared layout.

LouisFettet:

This is a follow-up to #273 and depends on work done in that PR.

## Proposed changes

- Many changes are from renaming / moving directories around:
  - `Breadcrumb` is now `Breadcrumbs`
  - Moved `/[topTierPage` and `/about` (and all their children) under the `/(infoPage)` shared layout so that we can set up breadcrumbs (and later the sidenav) in one place.
- Constructs breadcrumbs while we're building out the `pageMetadataMap` introduced in #273.
- Made some minor changes toe the `Breadcrumbs` component:
  - Removed pattern where the current page and the rest of the path are provided separately.
  - Used reactive values so that the component updates when the user navigates.
- Removed previous `Breadcrumbs` usage on the Office Page layout (moved into `(infoPage)` shared layout).

Only thing I can think of that might cause issues in the future is that the shared layout may need some variations to account for a hero image, but I figure we can cross that bridge when we come to it.

- [x] Updated breadcrumbs

Can be tested on any "info" page (feel free to click up the tree back to Home):
* [/animals/branding](https://ldaf-git-louis-nav-breadcrumbs-ldaf.vercel.app/animals/branding)
* [/land/fire/safety](https://ldaf-git-louis-nav-breadcrumbs-ldaf.vercel.app/land/fire/safety)
* [/about/organization/for](https://ldaf-git-louis-nav-breadcrumbs-ldaf.vercel.app/about/organization/for)


src/routes/loadPageMetadataMap.ts
```diff
@@ -60,6 +72,29 @@ const constructFullPathFromMap = (pageMetadata: PageMetadataMapItem, path = ""):
...
 
+// Recursive function that puts together breadcrumbs for a page.
+const constructBreadcrumbs = (
+  metadata: PageMetadataMapItem,
+  breadcrumbs: Breadcrumbs = []
+): Breadcrumbs => {
+  let link = metadata.url;
+  if (metadata.isRoot) {
+    link = "/";
+  }
```
hinzed1127, line 83: This doesn't seem strictly necessary, since the response I'm seeing for the root element has a `url` value of `"/"`. Is this just an additional safeguard, or is that value addition a more recent update?

hinzed1127 (approved):
One small clarifying comment, but otherwise looks solid!

src/routes/loadPageMetadataMap.ts
LouisFettet: Yup, you're absolutely right. This was likely leftover from a previous implementation. Good catch! I'll remove it.

PR #304, opened 2023-07-13
LDAF-350 Populate main menu using page metadata.

LouisFettet:

## Proposed changes

This PR is only focused on populating the main menu; there are no changes to the nav components or design (this will come later). Most of the changes are to one file, `src/lib/components/Header/Nav/Nav.server.ts`, where we're replacing our query for Draft Navigation content with one for Menu content. The main difference between these approaches is that Draft Navigation made editors explicitly provide all menus and menu children, whereas Menu just collects the Top Tier pages (so that we can order them appropriately) and from there we use the existing `PageMetadataMap` and the featured service groups on the Top Tier pages to construct the menus with very minimal editor interaction.

The structure for each menu (i.e. the children under Animals, Plants and crops, etc.) is:
* the first link is a link to the Top Tier page for that menu (this will eventually render with an image, and will be a standalone column if we're using a mega menu)
* the next set of links are links to Service Groups featured on the Top Tier page; this allows editors to add pages that are lower in the hierarchy (e.g. `/land/fire/safety`) or from a different Top Tier altogether 
* the last set of links are all pages that are in the "second tier" of the hierarchy (e.g. `/animals/branding`), sorted alphabetically by their title (if a page is both a featured service group and a second-tier, we don't repeat them here)

- [x] Update megamenu

### DIFF
--- a/src/routes/+layout.server.ts
+++ b/src/routes/+layout.server.ts
@@ -2,9 +2,12 @@ import { loadPageMetadataMap } from "./loadPageMetadataMap";
 import { loadSiteTitle } from "$lib/components/Header/Title/Title.server";
 import { loadMainNav, loadSecondaryNav } from "$lib/components/Header/Nav/Nav.server";
 
-export const load = async () => ({
-  pageMetadataMap: loadPageMetadataMap(),
-  siteTitle: loadSiteTitle(),
-  secondaryNavItems: loadSecondaryNav(),
-  navItems: await loadMainNav(),
-});
+export const load = async () => {
+  const pageMetadataMap = await loadPageMetadataMap();
+  return {
+    pageMetadataMap,
+    siteTitle: loadSiteTitle(),
+    navItems: loadMainNav(pageMetadataMap),
+    secondaryNavItems: loadSecondaryNav(),
+  };
+};
--- a/src/lib/components/Header/Nav/Nav.server.ts
+++ b/src/lib/components/Header/Nav/Nav.server.ts
@@ -1,39 +1,42 @@
 import gql from "graphql-tag";
 import { print as printQuery } from "graphql";
+
 import { CONTENTFUL_SPACE_ID, CONTENTFUL_DELIVERY_API_TOKEN } from "$env/static/private";
 import getContentfulClient from "$lib/services/contentful";
 import mainNavTestContent from "./__tests__/MainNavTestContent";
 import secondaryNavTestContent from "./__tests__/SecondaryNavTestContent";
-import type {
-  DraftNavigationLink,
-  DraftNavigationMenu,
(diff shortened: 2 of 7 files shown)


src/routes/+layout.server.ts
```diff
@@ -2,9 +2,12 @@ import { loadPageMetadataMap } from "./loadPageMetadataMap";
...
-  siteTitle: loadSiteTitle(),
-  secondaryNavItems: loadSecondaryNav(),
-  navItems: await loadMainNav(),
-});
+export const load = async () => {
+  const pageMetadataMap = await loadPageMetadataMap();
+  return {
+    pageMetadataMap,
+    siteTitle: loadSiteTitle(),
+    navItems: loadMainNav(pageMetadataMap),
```
hinzed1127, line 10: this is less expensive than the `loadPageMetadataMap` query/transformation, but as a follow-up in the same vein as comments on #273, should this get cached and the value for `navItems` stored in a cookie, as well?

hinzed1127 (approved):
Between this and #273, I feel good about the structure of the `pageMetadataMap`. Same questions about persist that data so it's not calculated on each load, but the logic of the actual data transformations feels sound to me.

My only additional follow-up: For pages that are getting populated in the main menu but are resulting in `404`s on the actual page, where is the disconnect? Is that a Contentful issue somewhere? It might be useful for logging purposes to have some additional context provided in all the various places we throw a 404 ([here](https://github.com/la-ldaf/ldaf-site/blob/main/src/routes/about/organization/%5Bslug%5D/%2Bpage.server.ts#L78), for example).

src/routes/+layout.server.ts
LouisFettet: Yeah, I think that ideally all of these data structures (the page metadata map, the main menu items, and the sidenavs map) should be cached in some way. They could easily get out-of-sync with each other if we don't update them all at once.

LouisFettet:
@hinzed1127 good question about 404s. This is a (hopefully) temporary problem that's arising when a `Page details` entry (artist previously known as `Page metadata`) is published, but its corresponding content, such as a `Core content` entry (artist previously known as `Service group`) is _not published_.  There's not a great way to determine if the link is going to lead to a 404 without doing some extra legwork to figure out the corresponding content type (could be top tier, core content, office page, board / commission etc.), so for now I just left this as-is. 

I say that this is temporary because all future content should require a `Page details` entry to be published alongside whatever content is being published, and since `Page details` shouldn't be showing up on their own in Compose, it should be way more unlikely to pop up. All that said, I do agree with you that it would be good to log this (like I'm already doing for pages where we can't resolve a path to the root), so I'll see about adding that in right quick.

PR #305, opened 2023-07-13
LDAF-350 Populate side navs using page metadata. 

LouisFettet:

This is the final PR focused on populating nav menus and links, and it depends on work done in #273, #293, and #304.

## Proposed changes

This PR is focused on populating the Side Navs that appear on pretty much every page except the homepage. We actually don't need to make any additional GraphQL queries to Contentful here, since the data we receive for building out the Page Metadata Map and the Main Menu give us everything we need to build the Side Navs.

Here's a quick breakdown of the changes:
- Move `<main>` element into the base `infoPages` layout to cut down on repetition now that we've handled all the major layout pieces (this results in very minor changes to a lot of `+page.svelte` files).
- Remove old `layout` files from the `officePage` since we're now handling everything in the `infoPages` layout.
- Build out another Map where the Top Tier slugs are the keys and the Side Nav trees are the values. This allows us to have one data structure we can pull any of the Side Navs from (e.g. one for `animals`, one for `plants`, etc.). Each item in a Side Nav tree has:
  - an `id`, useful for fetching Page Metadata and for SvelteKit to `key` on
  - a`title` for the links
  - a `link` for the actual URL to navigate to
  - `children` to allow nesting other Side Nav items appropriately
- Improve the `SideNav` component (and `SideNavItem`) so that we can:
  - add expand / collapse icons when an item has children
  - more easily determine when to highlight a link as "current" without having to track it anywhere (we just look at the URL)
- Add some styling on top of the base USWDS Side Nav to support expand / collapse icons.
- Make minor updates to corresponding test data, tests, and stories based on above changes to the `SideNav` component.

- [x] Updated side nav

With the Side Navs now in place, you should now be able to navigate to every page on the site that lives under a Top Tier page! Click around and you should see a stubbed page for every Page Details entry in Contentful (although some may lead to 404s if the Page Details are published but the corresponding content is not).


src/lib/components/SideNav/SideNav.server.ts
```diff
@@ -0,0 +1,84 @@
...
+      metadata.sys &&
+      metadata.title &&
+      metadata.url &&
+      topTier.children
+    ) {
+      // utilize the sort order from the main menu
+      const featuredPages = topTier.children.map((menuItem) => menuItem.id);
+      // remove the first item since it's always the top tier page
+      featuredPages.splice(0, 1);
+      sideNavMap.set(metadata.slug, {
```
LouisFettet, line 75: I figured it would be easier to use the metadata's `slug` instead of the metadata's `sys.id` for the keys in this map so that we could more easily get the appropriate sidenav from a nested page (easier to get the first slug in the URL than to traverse up the tree to get the metadata ID). If y'all have a strong preference here I can switch this to use ID instead.

src/lib/components/SideNav/SideNav.scss
```diff
@@ -0,0 +1,13 @@
...
+  .ldaf-sidenav-link-text {
+    width: 75%;
+  }
+  .ldaf-sidenav-icon-container {
+    width: 25%;
+    display: flex;
+    align-items: center;
+    justify-content: flex-end;
+  }
+}
```
LouisFettet, line 13: Styling here is just focused on properly rendering the collapse / expand icons next to links that have child pages.

LouisFettet (commented):
quick self-review

src/lib/components/SideNav/SideNav.server.ts
```diff
@@ -0,0 +1,84 @@
...
+const sideNavMap: SideNavMap = new Map();
+
+// recursive function that builds out a tree of children, grandchildren, etc.
+const buildSideNavTree = (
+  pageMetadataMap: PageMetadataMap,
+  featuredPages: string[],
+  metadata: PageMetadataMapItem
+): SideNavItem[] => {
+  const tree: SideNavItem[] = [];
+  const { children } = metadata;
```
hinzed1127, line 16: nit: if this is the only thing you need from `metadata`, why not just pass in `metadata.children` as the third argument to `buildSideNavTree`?

hinzed1127 (approved):
I think I've reached my saturation point on grokking recursive functions for the day, but this looks good! I think the next steps in my mind are:

1. getting all 4 parts of LDAF-350 merged in succession
2. fix anything that pops up in the merging
3. add in caching of pageMetadata

After seeing the full picture, (3) seems isolated enough that it shouldn't be a blocker to the rest of this going in, in my opinion. Honestly, just in case the caching itself has some bugs, it would be easier to revert/fix if its in its own PR, so I would probably prefer that.

src/lib/components/SideNav/SideNav.server.ts
LouisFettet: You're right, good catch! I'll make this change now.

LouisFettet: Knocked this out in 0ff1908!

PR #317, opened 2023-07-18
Create Service Group Page

hinzed1127:
Jira tickets: 
- [LDAF_332](https://ldaf.atlassian.net/browse/LDAF-332)
- Also knocked out [LDAF-290](https://ldaf.atlassian.net/browse/LDAF-290) (the Accordion component)

## Proposed changes

- Adds service group page via graphQL query and then conditionally rendering page contents based on the returned data
- Adds `Accordion` component that gets utilized by the Service Entry collection

## Screenshots

Easiest thing to do right now is navigate to the deployed site and the `/land/indian-creek` route

- [x] tests for `<Accordion />` 
- [x] matches the [Indian Creek Mockup](https://www.figma.com/file/JdZep2KGWtsAjqaHrfYoYk/LDAF-Prototypes?type=design&node-id=836-16586&mode=design&t=ukIbQxASnhakE4ZO-4) to a certain extent. There's gaps in what content I'm getting back from Contenful and information needed that will get addressed in follow-up work. Some of it is reliant on other open PRs (asset rendering, navigation, etc) 

## Other details
Overall diff is a little bloated right now `industrial-hemp-data.json` and  `pest-control-data.json`. That was the easiest way I found to quickly toggle between a few different service group pages while iterating and before the navigation/routing PRs get merged. They can get removed before merging.

## Requested feedback

- see self-review

[LDAF-290]: https://ldaf.atlassian.net/browse/LDAF-290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/routes/land/[slug]/+page.svelte
```diff
@@ -0,0 +1,93 @@
...
+{#if description}
+  <ContentfulRichText document={description?.json} />
+{/if}
+
+{#if serviceEntriesCollection}
+  <h2>{serviceListName}</h2>
+  {#if serviceEntries.length > 0}
+    {#each serviceEntries as item}
+      <Accordion multiselectable>
+        <AccordionItem title={item?.entryTitle}>
```
Comment (LouisFettet) [inline src/routes/land/[slug]/+page.svelte:63]: Might be good to provide `id` here as well with the Service Entry's `sys.id` so that we don't need to rely on the default `Math.random` value in the component.

src/routes/land/[slug]/+page.svelte
```diff
@@ -0,0 +1,93 @@
...
+  $: serviceEntries = [];
+  $: serviceGroups = [];
+
+  $: {
+    data.serviceEntriesCollection?.items.map((item) => {
+      if (item && "entryTitle" in item) {
+        serviceEntries?.push(item);
+      } else if (item && "title" in item) {
+        serviceGroups?.push(item);
+      }
```
Comment (LouisFettet) [inline src/routes/land/[slug]/+page.svelte:40]: A better way to determine whether the reference is a "service group" or "service entry" is to use the built-in `__typename` field; I did something similar here (because the reference could have been a "menu", "menu item", or "top tier"): 
https://github.com/la-ldaf/ldaf-site/blob/84a95a69ed8900f5e6fac5f3c74b0b5978b4cf1e/src/lib/components/Header/Nav/Nav.server.ts#L20
https://github.com/la-ldaf/ldaf-site/blob/84a95a69ed8900f5e6fac5f3c74b0b5978b4cf1e/src/lib/components/Header/Nav/Nav.server.ts#L65

If you do this in the `+page.server.ts` file you might not need the reactive statement here.

LouisFettet (commented):
This is looking good! Have a couple minor notes that will hopefully help somewhat.

src/routes/land/[slug]/+page.svelte
Comment (hinzed1127) [reply src/routes/land/[slug]/+page.svelte]: nice, I'm much more into that. I tried messing around with `__typename` but couldn't figure it out. this works, though, so I'm pushing an update in a similar vein

src/routes/land/[slug]/+page.svelte
Comment (hinzed1127) [reply src/routes/land/[slug]/+page.svelte:47]: Agreed. I'll keep the `Math.random` call in the component as a fallback in case `id` isn't supplied, but I added a clarifying comment and updated it to `sys.id` in this specific instance

src/lib/components/Accordion/AccordionItem.svelte
```diff
@@ -0,0 +1,52 @@
+<script lang="ts">
+  import { getContext, onMount } from "svelte";
+  import type { AccordionContext, AccordionContextItems } from "./types";
+
+  // Note: `id` has a fallback value, but ideally,
+  // one should supply their own identifiers
+  export let id = "accordionItem-" + Math.random().toString(36);
```
hinzed1127, line 7: I thought about just making this required, but I didn't want to necessarily enforce supplying your own `id`. I'd be happy to update this depending on what other's think.

hinzed1127 (commented):
Self-review

Most of it is with respect to the various WIP surrounding asset rendering and navigation/route generation that is encapsulated by other open PRs

src/lib/components/Accordion/AccordionItem.svelte
```diff
@@ -0,0 +1,52 @@
...
+      if (unsubscribe) unsubscribe();
+    };
+  });
+
+  $: if (context) {
+    context.add({ id, expanded });
+    unsubscribe = context.items.subscribe((value: AccordionContextItems) => {
+      expanded = value[id];
+    });
+  }
```
hinzed1127, line 28: Looking over this again, this seems like a danger of adapting others' code. This feels a bit convoluted and that I should just refactor this to use a store. Context, as I understand it, isn't reactive, whereas store _is_ and is meant to be subscribed to.

hinzed1127: So basically doing what [this article](https://joyofcode.xyz/svelte-context-with-stores) describes and using context to pass a store to descendant components. Then you get the clean API of stores with dealing with the global scope of a store

src/lib/components/Accordion/Accordion.svelte
```diff
@@ -0,0 +1,36 @@
...
+    toggle: (item: AccordionItemType) => {
+      items.update((context: AccordionContextItems) => {
+        if (!multiselectable) {
+          Object.keys(context).forEach((id) => (context[id] = false));
+        }
+
+        return { ...context, [item.id]: item.expanded };
+      });
+    },
+  });
```
Benaiah, line 31: A couple thoughts:

- `add` should be named `set`
- `set` should take two arguments, an ID and a boolean representing the expanded state
- `remove` should take an ID, not a whole `AccordionItemType`
- `toggle` should take an ID, not a whole `AccordionItemType`

src/lib/components/Accordion/AccordionItem.svelte
```diff
@@ -0,0 +1,52 @@
...
+  onMount(() => {
+    return () => {
+      if (context) context.remove({ id, expanded: !expanded });
+      if (unsubscribe) unsubscribe();
+    };
+  });
+
+  $: if (context) {
+    context.add({ id, expanded });
+    unsubscribe = context.items.subscribe((value: AccordionContextItems) => {
```
Benaiah, line 25: This reactive statement will rerun every time `expanded` changes, which means we will repeatedly subscribe to the context and throw away the old subscription. We should either call the old unsubscribe before reassigning it to the new subscription, or (preferably) refactor this so that the subscription doesn't depend on the value of `expanded`. Something like this should work (once other suggestions about the context methods are taken into account):

```ts
  const { items: accordionItems } = context;
  onMount(() => () => context && context.remove(id));
  $: if (context) context.set(id, expanded);
  $: if (context) expanded = $accordionItems[id] ?? false;
```

Using the auto-subscription like this should also allow us to omit the explicit `unsubscribe()` from the cleanup function returned by `onMount`.

src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.svelte
```diff
@@ -1,7 +1,79 @@
...
+{#if serviceEntries.length > 0 || serviceGroups.length > 0}
+  <h2>{serviceListName}</h2>
+  {#if serviceEntries.length > 0}
+    {#each serviceEntries as item}
+      <Accordion multiselectable>
+        <AccordionItem title={item?.entryTitle} id={item.sys.id}>
+          <ContentfulRichText document={item?.description?.json} />
+        </AccordionItem>
+      </Accordion>
+    {/each}
```
Comment (Benaiah) [inline src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.svelte:44]: Shouldn't the `<Accordion multiselectable>` tag be _outside_ of the `#each` block?

src/lib/components/Accordion/types.ts
```diff
@@ -0,0 +1,15 @@
...
+};
+
+export type AccordionContextItems = Record<string, boolean>;
+
+export type AccordionContext = {
+  items: Writable<AccordionContextItems>;
+  add: (item: AccordionItemType) => AccordionContext;
+  remove: (item: AccordionItemType) => AccordionContext;
+  toggle: (item: AccordionItemType) => AccordionContext;
+};
```
Benaiah, line 15:
```suggestion
export type AccordionContext = {
  items: Writable<AccordionContextItems>;
  set: (id: string, expanded: boolean) => AccordionContext;
  remove: (id: string) => AccordionContext;
  toggle: (id: string) => AccordionContext;
};
```

src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts
```diff
@@ -45,19 +126,38 @@ export const load = async ({ parent, params }) => {
...
+        ) as ServiceGroup[];
+
+        serviceGroups = serviceGroups.map((serviceGroup) => {
+          const serviceGroupMetadata = pageMetadataMap.get(
+            serviceGroup?.pageMetadata?.sys.id || ""
+          );
+          return { ...serviceGroup, url: serviceGroupMetadata?.url };
+        });
+
+        console.log("Matching metadata in service group's +page.server", matchedPageMetadata);
```
Comment (Benaiah) [inline src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts:154]: Looks like a debug statement got left behind here.

Benaiah (changes requested):
I think a refactor that made it so that all the state was stored in `Accordion` and we used events to bubble up toggling to the `Accordion` element, but this works fine and I think is worth merging once a few comments are addressed.

src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.svelte
Comment (hinzed1127) [reply src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.svelte:44]: yes! good catch

src/lib/components/Accordion/AccordionItem.svelte
```diff
@@ -0,0 +1,52 @@
...
+  import { getContext, onMount } from "svelte";
+  import type { AccordionContext, AccordionContextItems } from "./types";
+
+  // Note: `id` has a fallback value, but ideally,
+  // one should supply their own identifiers
+  export let id = "accordionItem-" + Math.random().toString(36);
+  export let title: string | null | undefined = "Title";
+  export let expanded = false;
+
+  const context: AccordionContext = getContext("Accordion");
```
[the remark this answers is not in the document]
hinzed1127: is that a concern since it's via Context? I.e., since `AccordionItem`s should always be getting the context of their parent `Accordion`, is there an actual danger of that?

src/lib/components/ContentfulRichText/ContentfulRichText.svelte
```diff
@@ -14,6 +14,6 @@
   }
 </script>
 
-{#each document.content as subNode}
+{#each document.content as subNode (crypto.randomUUID())}
```
hinzed1127, line 17: Ideally I would grab an ID straight from from contentful to use as an ID, but there's not one directly on the `document` property. I'm open to alternatives, but this has content at least updating properly now

src/lib/components/Accordion/Accordion.svelte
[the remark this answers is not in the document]
hinzed1127: I diverged pretty dramatically in my refactor and simplified it quite a bit, so the previous comment not longer applies as much

src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts
```diff
@@ -26,6 +106,7 @@ const query = gql`
 `;
 
 export const load = async ({ parent, params }) => {
+  if (!CONTENTFUL_SPACE_ID || !CONTENTFUL_DELIVERY_API_TOKEN) return ServiceGroupPageTestContent;
```
Comment (hinzed1127) [inline src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts:109]: @LouisFettet I saw that you removed most of the the test content in your nav work. Is this still needed?

src/lib/components/Accordion/Accordion.svelte
```diff
@@ -1,32 +1,23 @@
...
+    subscribe: selectedItems.subscribe,
+    toggle: (itemID: string) => {
+      if (multiselectable) {
+        selectedItems.update((items) => ({
+          ...items,
+          [itemID]: !items[itemID],
+        }));
+      } else {
+        selectedItems.update((selectedItems) => ({ [itemID]: !selectedItems[itemID] }));
+      }
```
Benaiah, line 20:
```suggestion
    toggle: (itemID: string) =>
      selectedItems.update((items) => ({
        ...(multiselectable ? items : {}),
        [itemID]: !items[itemID],
      })),
```

src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts
```diff
@@ -45,19 +122,37 @@ export const load = async ({ parent, params }) => {
...
-      if (matchedPageMetadata) {
-        return { pageMetadata: matchedPageMetadata };
+
+      if (matchedServiceGroup) {
+        const serviceEntries = matchedServiceGroup.serviceEntriesCollection?.items.filter(
+          (item) => item?.__typename === "ServiceEntry"
+        );
+        let serviceGroups = matchedServiceGroup.serviceEntriesCollection?.items.filter(
+          (item) => item?.__typename === "ServiceGroup"
+        ) as ServiceGroup[];
```
Comment (Benaiah) [inline src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts:141]: This is a small nit so it's not worth blocking the PR for, but for future reference it's safer to use a type generated from the query or explicitly defined where you need it rather than casting to a type from the schema, since the schema includes _all_ fields, even those that aren't included in the query.

src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts
```diff
@@ -45,19 +122,37 @@ export const load = async ({ parent, params }) => {
...
+          );
+          return { ...serviceGroup, url: serviceGroupMetadata?.url };
+        });
+
+        return {
+          ...matchedServiceGroup,
+          pageMetadata: matchedPageMetadata,
+          serviceEntries,
+          serviceGroups,
+        } as ServiceGroupPage;
```
Comment (Benaiah) [inline src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts:155]: Same nit as above.

Benaiah (approved):
Looks great, good to merge! I just have one optional suggestion and a note for future reference but no blockers.

src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts
Comment (hinzed1127) [reply src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts:141]: ah, that makes sense. I was trying to avoid casting, but resolving the TS errors without it was proving to be really difficult. I'm down to try to improve that aspect as follow-up

PR #330, opened 2023-07-24
Suppress @rollup/plugin-typescript warnings on +page.server.ts files.

LouisFettet:
The irrelevant `@rollup/plugin-typescript` warnings we're getting when running `dev` are coming from one of our custom Vite plugins, so for now we can just suppress the one that's causing the most false positives (the implicit type warning). In the future we may want to suppress all warnings coming from the TS plugin, since it doesn't have the full view of the application (nor should it).

### DIFF
--- a/vite-plugin-import-as-bundle-string/index.ts
+++ b/vite-plugin-import-as-bundle-string/index.ts
@@ -24,6 +24,13 @@ export default () => {
       const bundle = await rollup({
         input: path,
         plugins,
+        // Suppress some TS warnings that we know are irrelevant.
+        onwarn: (warning, handler) => {
+          const { pluginCode } = warning;
+          // TS7031 - https://github.com/microsoft/TypeScript/blob/cbf3c63ef3bb85e235092eaf5aa6035dad04b185/src/compiler/diagnosticMessages.json#L6382-L6385
+          if (pluginCode === "TS7031") return;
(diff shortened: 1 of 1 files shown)


Benaiah:
IMO we should figure out a way to disable these particular rollup warnings instead. `npm run check` is already type-checking these load functions (and we get LSP autocomplete for them), which you can confirm by using a nonexistent piece of data and seeing the type error. It seems like SvelteKit has a TypeScript plugin that works in LSP and when running `npm run check` but rollup doesn't know about it.

LouisFettet:
@Benaiah before I update the PR description, what do you think of the new solution (force-pushed a new commit)? Right now we're just suppressing the one TS warning that seems to be the peskiest, but we could just suppress all TS warnings coming from this plugin if we wanted to.

Benaiah:
> @Benaiah before I update the PR description, what do you think of the new solution (see commit [531b390](https://github.com/la-ldaf/ldaf-site/commit/531b3906aaba5c37b718ea25c005d884b08f77ce))? Right now we're just suppressing the one TS warning that seems to be the peskiest, but we could just suppress all TS warnings coming from this plugin if we wanted to.

This looks good to me! We can always suppress more stuff later if it continues to annoy.

hinzed1127 (approved):

PR #339, opened 2023-07-26
LDAF-348 Create VideoCard component.

LouisFettet:
Jira ticket: [LDAF-348](https://ldaf.atlassian.net/browse/LDAF-348)

## Proposed changes

- Fetch data from YouTube API for the video title and description when `customTitle` and `customDescription` are not provided.
- Add a YouTube subscribe button.
- Add styling and fix some minor spacing issues on the standard cards on the Top Tier and Service Group pages.
- Add basic tests and story.

## Requested feedback

See self-review.

[LDAF-348]: https://ldaf.atlassian.net/browse/LDAF-348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/lib/components/VideoCard/VideoCard.svelte
```diff
@@ -17,17 +34,44 @@
...
-    title="Embedded YouTube video"
-    frameborder="0"
-    allowfullscreen
-  />
+  {#key youtubeVideoId}
+    <div class="ldaf-video-card ldaf-video-card--{variation} {className}">
+      <div class="ldaf-video-container">
+        <iframe
+          class="ldaf-video-embed"
+          src={`https://www.youtube-nocookie.com/embed/${youtubeVideoId}`}
```
LouisFettet, line 53: Using the `nocookie` url is supposed to disable trackers (until the user actually watches a video).

src/lib/components/VideoCard/VideoCard.svelte
```diff
@@ -17,17 +34,44 @@
...
 {#if youtubeVideoId}
-  <iframe
-    width="560"
-    height="315"
-    src={`https://www.youtube.com/embed/${youtubeVideoId}`}
-    title="Embedded YouTube video"
-    frameborder="0"
-    allowfullscreen
-  />
+  {#key youtubeVideoId}
```
LouisFettet, line 48: Mentioned this before, if we don't include this then the "subscribe" button below gets re-added for some reason. There's probably a better way to handle this so that both the `afterUpdate` and the `key` aren't needed.

src/variables.scss
```diff
@@ -32,7 +32,7 @@ $accent-dark: #ffb200;
 $accent-darker: #d49400;
 
 $grayscale-01: #f9f9f9;
-$grayscale-05: #f2f2f0;
+$grayscale-05: #f0f0f0;
```
LouisFettet, line 35: There was a typo in the Figma file; this is the correct value.

LouisFettet (commented):
self-review

Benaiah (approved):
Looks good to me! I'd like to take another look at the styling, the issue that's making us use the `key` block, and fetching the video data on the server side so it's present in the initial response, but none of that is urgent enough to block this. The only issue I noticed is this visual bug at small mobile sizes - looks like the video iframe is not resizing to fill its container if the container is too small.

image

src/lib/components/VideoCard/VideoCard.svelte
```diff
@@ -17,17 +34,43 @@
       console.warn("Unsupported video type; could not embed video.");
     }
   }
+
+  afterUpdate(async () => {
+    if (youtubeVideoId && (!title || !description)) {
+      const snippet = await getYoutubeVideoData(youtubeVideoId);
+      title = snippet.title;
+      description = snippet.description;
```
Benaiah, line 42: Small nit that doesn't need to block the PR: I think it would be cleaner to make these separate variables instead of having the component reassign its own props.

LouisFettet:
@Benaiah thanks for the review. I thought that the video not shrinking below `300px` was an issue with the embed, but it turns out it was an issue with my flexbox rules. This should be fixed. I also added reactive values for `title` and `description` so we're not resetting props. Let me know what you think! I'll probably merge this first thing tomorrow if this looks good to you.

Benaiah (approved):
Looks great!

PR #347, opened 2023-08-02
Create new version of Home Page with content from Contentful

LouisFettet:
Jira ticket: [LDAF-342](https://ldaf.atlassian.net/browse/LDAF-342)

## Proposed changes

Adds a rough draft of the Home Page, grabbing the embed video data and the featured `Core Content` pages from Contentful.

- Rewrites the base `+page.server.ts` and `+page.svelte` files for new content. Heavily utilizes patterns introduced by @Benaiah in #279.
- Makes a handful of fixes to the `VideoCard` component that I missed in initial implementation.
- Adds a `ConditionalWrapper` component that allows us to conditionally wrap HTML and Svelte components in an HTML element if the condition is met. Used on the site title on the home page to guarantee we have an `h1`. 
- Removes the test data and images previously used on the Home Page, as well as the old styling.
- Fixes end-to-end tests.

## To-Do (in follow-ups)

- Shift headings on home page to `h2` instead of `h3` for semantic HTML ([LDAF-369](https://ldaf.atlassian.net/browse/LDAF-369)).
- Handle proper layout of cards so that the first is full-width, the second and third are half-width, and all following are third-width. Also need to make some changes so that the images in the cards take up more space when the cards are larger ([LDAF-370](https://ldaf.atlassian.net/browse/LDAF-370)).

- [x] Video componen
- [x] Featured services displayed as cards

## Requested feedback

See self-review.

[LDAF-342]: https://ldaf.atlassian.net/browse/LDAF-342?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[LDAF-369]: https://ldaf.atlassian.net/browse/LDAF-369?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LDAF-370]: https://ldaf.atlassian.net/browse/LDAF-370?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/lib/components/VideoCard/VideoCard.svelte
```diff
@@ -37,43 +37,55 @@
...
 
   afterUpdate(async () => {
     if (youtubeVideoId && (!customTitle || !customDescription)) {
       const snippet = await getYoutubeVideoData(youtubeVideoId);
-      title = snippet.title;
-      description = snippet.description;
+      if (!customTitle) {
+        title = snippet.title;
+      }
+      if (!customDescription) {
```
LouisFettet, line 48: Previously both the `customTitle` and `customDescription` were getting overwritten if one was blank. This way we only overwrite the blank ones.

src/lib/components/VideoCard/VideoCard.svelte
```diff
@@ -37,43 +37,55 @@
...
-          class="ldaf-video-embed"
-          src={`https://www.youtube-nocookie.com/embed/${youtubeVideoId}`}
-          title="Embedded YouTube video player"
-          frameborder="0"
-          allowfullscreen
+  <div class="ldaf-video-card ldaf-video-card--{variation} {className}">
+    <div class="ldaf-video-container">
+      <iframe
+        class="ldaf-video-embed"
+        src={`https://www.youtube-nocookie.com/embed/${youtubeVideoId}?rel=0`}
```
LouisFettet, line 64: Addition of the `rel=0` makes it so that YouTube only recommends other videos by LDAF when the user pauses.

src/lib/components/VideoCard/VideoCard.svelte
```diff
@@ -37,43 +37,55 @@
...
+        title="Embedded YouTube video player"
+        frameborder="0"
+        allow="encrypted-media;"
```
LouisFettet, line 67: Adding this resolved some console errors I was getting. There's a lot more options that YouTube provides by default but I wanted to keep the permissions here slim.

src/routes/+page.svelte
```diff
@@ -1,79 +1,82 @@
...
+  import Image from "$lib/components/Image/Image.svelte";
 
   export let data;
+  $: ({
+    homePage: { heroVideo, featuredServices },
+  } = data);
+
+  const getCardSettings = (
+    index: number
+  ): { card: "full" | "half" | "third"; button: Variant; imageLoading: "lazy" | "eager" } => {
```
LouisFettet, line 21: For the first draft we're not doing anything fancy with the card layout, but I figured it would be useful later.

LouisFettet (commented):
self-review

src/routes/+page.svelte
```diff
@@ -1,79 +1,82 @@
...
+        {@const { card: cardSize, button: buttonVariant, imageLoading } = getCardSettings(index)}
+        <!-- TODO: Can't conditionally render a named slot, but ideally we only declare Card once here. -->
+        {#if item?.heroImage?.imageSource?.url}
+          <Card class={`usa-card--${cardSize}`}>
+            <h3 class="usa-card__heading" slot="header">{item.title}</h3>
+            <Image
+              slot="image"
+              src={item.heroImage.imageSource.url}
+              alt={item.heroImage.imageSource.title || "Card image"}
+              sources={getSources(item.heroImage.imageSource.url)}
```
Benaiah, line 50: This fixes the images disappearing at large sizes. The problem was that `getSources` wasn't being told about the image's width so it generated sizes that were larger than the image itself. `sources` is only supposed to be an array when you have a static set of images (like our build-time assets) - if you have a URL generator like `getSources` you can pass that URL generator into the `Image` component as `sources` and it will automatically call it with the image's metadata (currently just height and width) and use that as the sources.

```suggestion
              sources={getSources}
```

src/routes/+page.svelte
```diff
@@ -1,79 +1,82 @@
...
+              blurhash={item.heroImage.imageSource?.blurhash ?? undefined}
+              width={item.heroImage.imageSource.width ?? undefined}
+              height={item.heroImage.imageSource.height ?? undefined}
+              fit
+              loading={imageLoading}
+            />
```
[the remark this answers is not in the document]
Benaiah: It's a one-line fix; I've suggested the change to fix it above.

Benaiah (approved):
Looks great! The image issue you noticed is a one-line fix, and I've suggested the change in my review.

Benaiah:
Oh, I forgot to note something - once https://github.com/la-ldaf/ldaf-site/pull/348 is merged, we should change the card images on the homepage to specify `sizeType="card"`. Right now they're generating sizes designed for full-bleed images (like the hero image). Correctly sizing them should make the homepage a lot smaller.

src/routes/+page.svelte
LouisFettet: Thanks for the assist here!

src/lib/assets/under-construction.gif
hinzed1127, line 1: RIP early-2000s Web 1.0 aesthetic 🪦

src/lib/components/ConditionalWrapper/ConditionalWrapper.svelte
hinzed1127, line 1: This is clever. Thinking out loud, are there other instances in which we think we'd want to utilize this?

src/routes/+page.svelte
```diff
@@ -1,79 +1,83 @@
 <script lang="ts">
-  import "./page.scss";
-  import ContentfulRichText from "$lib/components/ContentfulRichText";
-  // import VideoCard from "$lib/components/VideoCard";
+  import "./(infoPages)/[topTierPage]/page.scss";
```
hinzed1127, line 2: Follow-up work: another point in favor of doing the `TODO` in that file (create a `<CardGroup />` wrapper component) now that we've got in a few different places

hinzed1127 (approved):
🚀 lookin good

src/lib/components/ConditionalWrapper/ConditionalWrapper.svelte
LouisFettet: Wish I had thought of it myself! I had run into a similar use-case earlier when I thought that we might have to create a new container for the VideoCard on the home page, so I implemented it and then removed it after talking with Courtney. I'm sure we'll run into needing this somewhere else down the line.

PR #348, opened 2023-08-02
Use Image component instead of img tags, reset image state on src change

Benaiah:
## Proposed changes

- Uses `Image` for the hero image and card images on top tier pages
- Uses a very small version of the logo to display in the banner
- Resets the state of the Image component on src change, fixing the stale-image-on-client-side-navigation issue
- Some small fixes

- [x] Tested manually
- [x] All tests continue to pass

### DIFF
--- a/src/routes/(infoPages)/+layout.svelte
+++ b/src/routes/(infoPages)/+layout.svelte
@@ -31,6 +31,7 @@
       blurhash={topTierPage.heroImage.imageSource.blurhash}
       fit={true}
       preserveAspectRatio={false}
+      loading="eager"
     />
     {#if topTierPage.heroImage.fotogCredit}
       <div class="grid-container">
--- a/src/routes/(infoPages)/+layout.svelte
+++ b/src/routes/(infoPages)/+layout.svelte
@@ -1,8 +1,11 @@
 <!-- Shared layout for all informative pages (Top Tiers, Service Groups, Offices, Boards/Commissions)-->
 <script lang="ts">
+  import "./layout.scss";
   import { page } from "$app/stores";
   import Breadcrumbs from "$lib/components/Breadcrumbs";
   import SideNav from "$lib/components/SideNav";
+  import Image from "$lib/components/Image";
+  import { getSources } from "$lib/imageServices/contentful";
 
   $: ({ pathname } = $page.url);
   $: ({ pageMetadata, sideNavMap, __typename, topTierPage } = $page.data);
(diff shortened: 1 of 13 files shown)


src/routes/(infoPages)/+layout.svelte
```diff
@@ -16,20 +19,21 @@
...
+      src={topTierPage.heroImage.imageSource.url}
+      alt={topTierPage.heroImage.imageSource.description}
+      sources={getSources}
+      width={topTierPage.heroImage.imageSource.width}
+      height={topTierPage.heroImage.imageSource.height}
+      sizeType="full-bleed"
+      blurhash={topTierPage.heroImage.imageSource.blurhash}
+      fit={true}
+      preserveAspectRatio={false}
+      loading="eager"
```
LouisFettet, line 34: Most of these look good, but the one for `About LDAF` seems to be failing to load on larger screen sizes: https://ldaf-git-benaiah-use-image-component-instead-of-img-tags-ldaf.vercel.app/about

Might be an issue with the image itself?

LouisFettet (commented):
Looking good, thanks for updating the `load` function on the Top Tier pages. There's one Top Tier hero image that looks like it's failing to load, but otherwise everything looks good.

You might need to make some updates to the images in the cards on the Home Page now that that's been merged.

Also, if you have a second, this might be a good time to look into why the logo in the header gets squished on mobile (but feel free to save this for later if it's not a quick fix): 
Screenshot 2023-08-03 at 12 29 11 PM

src/routes/(infoPages)/+layout.svelte
Benaiah: It's another case of us trying to convert to AVIF when the image is over 9 megapixels. Just pushed a fix.

src/routes/+page.svelte
```diff
@@ -53,7 +53,7 @@
               blurhash={item.heroImage.imageSource?.blurhash ?? undefined}
               width={item.heroImage.imageSource.width ?? undefined}
               height={item.heroImage.imageSource.height ?? undefined}
-              fit
+              sizeType="card"
```
LouisFettet, line 56: Thanks for adding this! Can you remove the `TODO` comment before you merge?
https://github.com/la-ldaf/ldaf-site/blob/da1ce05ce0df3677a08272abb7bb628e76fb5087/src/routes/%2Bpage.svelte#L47

LouisFettet (approved):
LGTM! Have a very very minor comment but otherwise this is good to go!

PR #349, opened 2023-08-02
Fix service group page by splitting query

Benaiah:
## Proposed changes

- Splits up the service group page GraphQL query into separate queries for the base data and its children (service entries and groups).
- If more than 10 child service entries are present, chunks the queries into batches of 10 IDs at a time to avoid query complexity limits

- [x] Manually tested
- [x] All tests pass

### DIFF
--- a/src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts
+++ b/src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts
@@ -138,6 +138,9 @@ const childServiceGroupsQuery = gql`
   query ServiceGroupChildGroups($ids: [String]!) {
     serviceGroupCollection(limit: 10, where: { sys: { id_in: $ids } }) {
       items {
+        sys {
+          id
+        }
         pageMetadata {
           sys {
             id
@@ -195,6 +198,18 @@ export type ServiceGroupPage = {
   pageMetadata?: ServiceGroupMetadata;
 };
 
+const inOrder = <T>(items: T[], fn: (item: T) => string, order: string[]) => {
+  if (order.length !== items.length) {
+    throw new Error("ID order array does not match the provided items");
+  }
+  const ids = items.map(fn);
+  const record = Object.fromEntries(ids.map((id, i) => [id, items[i]]));
+  if (!order.every((id) => record[id])) {
+    throw new Error("ID order array does not match the provided items");
+  }
+  return order.map((id) => record[id]);
+};
+
 export const load = async ({
   parent,
   params: { topTierPage, serviceGroupPage },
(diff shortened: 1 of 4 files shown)


src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts
```diff
@@ -185,46 +213,89 @@ export const load = async ({
...
+    const childServiceGroupIDs =
       serviceGroup.serviceEntriesCollection?.items
-        .filter((item): item is ChildServiceGroup => item?.__typename === "ServiceGroup")
-        .map((group) => {
-          const { id } = group?.pageMetadata?.sys ?? {};
-          if (!id) return group;
-          const { url } = pageMetadataMap.get(id) ?? {};
-          return { ...group, url };
-        }) ?? [];
+        ?.filter((item): item is ChildServiceGroupStub => item?.__typename === "ServiceEntry")
```
Comment (LouisFettet) [inline src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts:238]: Looks like the cards aren't showing up on Core Content pages; I think this needs to be:
```suggestion
        ?.filter((item): item is ChildServiceGroupStub => item?.__typename === "ServiceGroup")
```

LouisFettet (changes requested):
Looks mostly good, but there's a bug I noted below where we're not filtering properly and thus aren't rendering the cards.

src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts
```diff
@@ -185,46 +213,89 @@ export const load = async ({
...
-    const data = await client.fetch<ServiceGroupCollectionQuery>(printQuery(query), {
+
+    const baseData = await client.fetch<ServiceGroupQuery>(printQuery(baseQuery), {
       variables: { metadataID },
     });
-    if (!data) break fetchData;
-    const [serviceGroup] = data?.serviceGroupCollection?.items ?? [];
+    if (!baseData) break fetchData;
+    const [serviceGroup] = baseData?.serviceGroupCollection?.items ?? [];
     if (!serviceGroup) break fetchData;
```
Comment (hinzed1127) [inline src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts:222]: since serviceGroup is always coalescing to `[]` when `undefined` or `null`, won't this this `break` statement never occur?

hinzed1127 (approved):
clarifying question on a possibly redundant line but otherwise good. This should probably go in before #350, as it will need to incorporate the split queries, as well as the additions it makes to the Service Page queries to grab entry links (in addition to assets)

src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts
Comment (Benaiah) [reply src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts:222]: The destructuring assignment here assigns `serviceGroup` to the first element of `baseData?.serviceGroupCollection?.items ?? []`. If `baseData?.serviceGroupCollection?.items` is `null` or `undefined`, `serviceGroup` will also be `undefined` because it's assigned to the first element of an empty array.

Comment (hinzed1127) [reply src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts:222]: got it. missed the destructuring bit

LouisFettet:
@Benaiah one other issue, it seems like we're no longer respecting the sort order of the services (both other groups and entries) from Contentful.

Examples:
* ["Selling food" Contentful entry](https://app.contentful.com/spaces/pc5e1rlgfrov/environments/master/entries/76gI1YBL6k4i2Dx9WZgQCO) and [`/food/selling`](https://ldaf-git-benaiah-fix-service-group-page-by-splitting-query-ldaf.vercel.app/food/selling) have different orders for the cards.
* ["Milk and dairy" Contentful entry](https://app.contentful.com/spaces/pc5e1rlgfrov/environments/master/entries/6U1CiXMrt5jxvAM26xaspZ) and [`/food/selling/dairy`](https://ldaf-git-benaiah-fix-service-group-page-by-splitting-query-ldaf.vercel.app/food/selling/dairy) have different orders for the accordions.

Benaiah:
@LouisFettet thanks for the catch; I've fixed the PR to preserve the authored order.

LouisFettet (approved):
LGTM! Thanks for fixing those issues. :shipit:

PR #350, opened 2023-08-03
Render CTAs and Contact info in Service Entries, Add more rich text node support

hinzed1127:
Jira tickets:
- [LDAF-359](https://ldaf.atlassian.net/browse/LDAF-359)
- [LDAF-359](https://ldaf.atlassian.net/browse/LDAF-365)

## Proposed changes

- Render Call To Action buttons in Service Entries
- Render Contact cards in Service Entries
- Add support for Embedded entry blocks, currently:
    - Contact
    - ImageWrapper
    - (still need future support for Location and VideoWrapper)

## Screenshots
Screenshot 2023-08-02 at 9 52 32 PM

- [x] Add and successfully render missing Rich Text nodes with links/entries (`AssetHyperlink`, `EmbeddedEntryBlock`, and `EmbeddedEntry`)
- [x] Core Content Pages include call to actions and contact information in Service entries (see screenshot above)

## Other details

Embedded Entries are a little less straightforward than other embeds, because they can take on a number of different types of entries. This involved more conditional logic in `EmbeddedEntryBlock` and `EmbeddedEntryBlock` to render the proper content based on their type.

### Possible drawbacks

One of the possible linked entries to Entry hyperlinks can be other pages. However, the response data for such entries is just the entry ID, so linking to the page requires retrieving that same ID from the metadata map. As such, the `pageMetadataMap` is added to the context of the `ContentfulRichText` component, since we have no way of predicting where in the component hierarchy that data might need to be accessed. This is a less of a drawback but more of a reality of the data model as it presently is (given how embedded content works in Contentful's response data, I'm not sure if there's an immediately better solution ). @Benaiah moved the page metadata loading under `$lib`, so it's at least a cleaner import now wherever it gets used.

[LDAF-359]: https://ldaf.atlassian.net/browse/LDAF-359?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LDAF-359]: https://ldaf.atlassian.net/browse/LDAF-359?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.svelte
```diff
@@ -48,43 +60,67 @@
...
-{#if childServiceGroups.length > 0}
-  <ul class="service-group-list">
-    {#each childServiceGroups as item}
-      <Card class="service-group-card">
-        <h3 class="usa-card__heading" slot="header">{item.title}</h3>
-        <svelte:fragment slot="body">
-          {#if item.subheading}
-            {item.subheading}
+          <!-- .filter is to make sure there aren't any null 
+            contact items(caused by contacts in a draft state) -->
```
Comment (hinzed1127) [inline src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.svelte:93]: this was caused by one of the service entries on `/about/contact/complaints`. there's a contact that's still in a draft state and the response for `items` was `[null]`

src/lib/fragments/entryProps.ts
```diff
@@ -0,0 +1,51 @@
...
+      phoneExt
+      email
+      # location {
+      #   name
+      #   streetAddress1
+      #   streetAddress2
+      #   city
+      #   state
+      #   zip
+      # }
```
hinzed1127, line 48: strangely, this bit was working fine in the Contentful GraphiQL IDE, but I was getting this error in the GraphQL validation in local development: `Unable to find field "location" on type "Contact"!`

hinzed1127 (commented):
Tiny bit of self-review

src/lib/fragments/entryProps.ts
Benaiah: Do you need to do `npm run gen-schema` maybe?

hinzed1127: hmmm that's giving its own problems:
`Query cannot be executed. Requested environment does not exist in the space`. Am I missing something being up-to-date?

I thought all the schema updating happened in other scripts run before `dev`. Strange that this one issue is occurring.

src/lib/components/ContentfulRichText/nodes/EntryHyperlink.svelte
```diff
@@ -19,12 +19,10 @@
...
-  let entryMetadata: PageMetadataMapItem;
-  if (entry?.__typename === "PageMetadata") {
-    const entryMetadata = linksContext.pageMetadataMap.get(entryID);
-    if (!entryMetadata) {
-      throw new Error(`the entry asset ${entryID} was not found in the pageMetadataMap`);
-    }
+  const entryMetadata =
+    entry?.__typename === "PageMetadata" && linksContext.pageMetadataMap.get(entryID);
+  if (!entryMetadata) {
+    throw new Error(`the entry asset ${entryID} was not found in the pageMetadataMap`);
```
hinzed1127, line 24: Reverting because this will throw errors for entries with `entry.__typename === "Contact"`

src/lib/components/ContentfulRichText/nodes/EntryHyperlink.svelte
```diff
@@ -0,0 +1,39 @@
...
+  if (entry?.__typename === "PageMetadata") {
+    const entryMetadata = linksContext.pageMetadataMap.get(entryID);
+    if (!entryMetadata) {
+      throw new Error(`the entry asset ${entryID} was not found in the pageMetadataMap`);
+    }
+  }
+</script>
+
+{#if entry?.__typename === "PageMetadata"}
+  <a href={entryMetadata?.url}
```
LouisFettet, line 32: nit: Use `Link` component. Will help us out in the future (we'll probably want to determine whether `rel="external"` programmatically to start indiicating when a user is leaving the site).

src/lib/components/ContentfulRichText/nodes/Node.svelte
```diff
@@ -7,19 +7,8 @@
   const component = nodes[node.nodeType];
 
   if (!component) {
-    // throw new Error(`unsupported nodeType ${node.nodeType}`);
+    throw new Error(`unsupported nodeType ${node.nodeType}`);
```
LouisFettet, line 10: Throwing an error here will cause a `500` yeah? Any chance we could log a warning instead and keep the if / else block below, rendering nothing if it's not supported? I'm still not sure if we've battened down the hatches on all the possible allowed rich text field references.

LouisFettet (approved):
Looks great, just have a couple notes which shouldn't require re-review in order to merge. Thanks for y'all's work on this! So exciting to see like 99% of the content getting rendered on the site!

src/lib/components/ContentfulRichText/nodes/Node.svelte
hinzed1127: good call, I'll move this to a `console.warn`

Benaiah (approved):

PR #352, opened 2023-08-03
LDAF-360 Create and implement Footer

LouisFettet:
Jira ticket: [LDAF-360](https://ldaf.atlassian.net/browse/LDAF-360)

## Proposed changes

- Adds new Footer component, styling, story, and tests. See [Figma mock](https://www.figma.com/file/oGKbyCnCRRdNzLYbiags93/LDAF-Component-Library-USWDS-3.0.2?type=design&node-id=2196-7873&mode=design&t=DmtzRj01Ub5sjpBR-4). Also includes logic to:
  - render menu headings as buttons on mobile (not a fan of this, but it's how the USWDS does it)
  - expand / collapse footer menus on mobile
  - collapse all menus on route change
- Fetch footer menu content from Contentful in a similar way to how we fetch the header menu content. Tweak some naming in the base layout server file to differentiate between nav items in the header versus in the footer.
- Make some minor adjustments to the base layout, including a target for the "Back to top" link.
- Stub out the Events and News aggregation pages.
- Make minor fix to end-to-end tests.

## Requested feedback

See self-review.

[LDAF-360]: https://ldaf.atlassian.net/browse/LDAF-360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/lib/components/Footer/Footer.scss
LouisFettet, line 1: Really not happy with the amount of hacking I had to do to get this working. Overriding colors is one thing but I had to shift the layout rules a good bit which makes this whole file a bit messy. Let me know if you see any glaring issues.

src/lib/components/Footer/Footer.svelte
```diff
@@ -0,0 +1,164 @@
...
+
+  export let navItems: NavMenuType[] = [];
+  export let siteTitle: SiteTitleType;
+  // we provide this from a parent component so that we can render the component in Storybook
+  export let afterNavigate: undefined | ((callback: (navigation: AfterNavigate) => void) => void) =
+    undefined;
+
+  // Quickest way to implement the accordions was to replicate what USWDS was
+  //   doing, which is to use buttons on mobile and headings on larger screens.
+  const MOBILE_BREAKPOINT = 480;
```
LouisFettet, line 23: Do y'all know if there's a way we can grab this breakpoint from USWDS (ideally without introducing a new dependency)?

LouisFettet (commented):
Quick self-review.

src/lib/components/Footer/Footer.svelte
```diff
@@ -0,0 +1,164 @@
...
+
+  const toggle = (i: number) => {
+    expandedIndex = expandedIndex === i ? undefined : i;
+  };
+
+  const handleKeyDown = (event: KeyboardEvent, i: number) =>
+    (event.key === "Enter" || event.key === " ") && toggle(i);
+  const handleMouseDown = (i: number) => toggle(i);
+
+  if (afterNavigate) afterNavigate(() => (expandedIndex = undefined));
```
LouisFettet, line 38: I would have preferred to use `navigating` (from `$app/stores`) but that gets set to true as soon as the user begins navigation, which seemed a bit disorienting (user would click, things would shift due to collapsing menus, then page would load).

Benaiah (approved):
LGTM!

PR #354, opened 2023-08-04
Remove dot-gov banner

hinzed1127:
Not removing the full banner in case there's a need at some point, but removing it from the actual header.


LouisFettet (approved):
LGTM

Benaiah (approved):

PR #355, opened 2023-08-04
fix EntryHyperlink

Benaiah:
## Proposed changes

- Fixes a bug in the `EntryHyperlink` component.

### DIFF
--- a/src/lib/components/ContentfulRichText/nodes/EntryHyperlink.svelte
+++ b/src/lib/components/ContentfulRichText/nodes/EntryHyperlink.svelte
@@ -5,7 +5,7 @@
   import { getContext } from "svelte";
   import { linksKey, type LinksContext } from "../context";
   import { isEntryHyperlink } from "../predicates";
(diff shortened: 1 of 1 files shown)


hinzed1127 (approved):

src/lib/components/ContentfulRichText/nodes/EntryHyperlink.svelte
```diff
@@ -16,13 +16,14 @@
 
   const linksContext = getContext<LinksContext | undefined>(linksKey);
   if (!linksContext) throw new Error("no context was provided for entry hyperlink");
+
   const { id: entryID } = entryHyperlink.data.target.sys;
 
   const entry = linksContext.linksEntriesMaps.hyperlink.get(entryID);
 
   let entryMetadata: PageMetadataMapItem;
```
hinzed1127, line 24: This is one way to clear up the type error
```suggestion
  let entryMetadata: PageMetadataMapItem | undefined;
```

PR #357, opened 2023-08-04
MVP Search with LDAF site metadata

hinzed1127:
Jira ticket: [LDAF-361](https://ldaf.atlassian.net/browse/LDAF-361)

## Proposed changes

- Adds server endpoint for requesting. a version of the pageMetadataMap
   - converts the map to an array of values for uploading to Algolia
   - omits breadcrumbs and adds in `metaTitle` and `metaDescriptions`
- Reveals `Search` component in the header
- Connects search to the proper [LDAF index](https://dashboard.algolia.com/apps/9IIDYROXZ5/explorer/browse/prod_ldaf?searchMode=search)

## Screenshots

Screenshot 2023-08-04 at 4 52 10 PM

- [x]  Simple search with site metadata

## Requested feedback

[LDAF-361]: https://ldaf.atlassian.net/browse/LDAF-361?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/routes/search/searchHelpers.ts
```diff
@@ -22,21 +22,22 @@ const getHighlightOrSnippetResultValue = (
...
+  // Docs: https://www.algolia.com/doc/api-reference/api-parameters/searchableAttributes/
+  return `
     <div>
       <article>
         <div class="post-content">
           <h2 class="entry-title">
-            <a class="usa-link" href="${hit.permalink}" rel="bookmark">
-              ${getHighlightOrSnippetResultValue(hit._highlightResult?.post_title)}
+            <a class="usa-link" href="${hit.url}" rel="bookmark">
+              ${getHighlightOrSnippetResultValue(hit._highlightResult?.metaTitle)}
```
hinzed1127, line 34: I'm guessing we'll be able to remove the wrapper function and just use type assertions. I just didn't get to that yet.

src/routes/api/v1/algolia/+server.ts
hinzed1127, line 1: My intention behind this had a few thoughts:

1. Reusing the pageMetadata logic as a single source of truth
2. If we want to have a separate webhook or other sort of scheduled job to update index values, this seems like a viable way to do that (especially for future changes where we'll probably be adding more than just the metadata map to the index

hinzed1127 (commented):
self-review

src/lib/loadPageMetadataMap.ts
```diff
@@ -97,7 +100,9 @@ const constructBreadcrumbs = (
   return breadcrumbs;
 };
 
-export const loadPageMetadataMap = async (): Promise<{
+export const loadPageMetadataMap = async (
+  includeBreadcrumbs = true
+): Promise<{
```
Benaiah, line 105:
```suggestion
export const loadPageMetadataMap = async ({ includeBreadcrumbs = true } = {}): Promise<{

```

src/routes/api/v1/algolia/+server.ts
```diff
@@ -0,0 +1,10 @@
+import { json } from "@sveltejs/kit";
+import type { RequestHandler } from "@sveltejs/kit";
+import { loadPageMetadataMap } from "$lib/loadPageMetadataMap";
+
+export const GET = (async () => {
+  const includeBreadcrumbs = false;
+  const { pageMetadataMap } = await loadPageMetadataMap(includeBreadcrumbs);
```
Benaiah, line 7:
```suggestion
  const { pageMetadataMap } = await loadPageMetadataMap({ includeBreadcrumbs: false });
```

Benaiah (approved):
Looks great! Can we make `includeBreadcrumbs` in `loadPageMetadataMap` a key on an options parameter tho? Positional boolean arguments are hard to keep track of so it's nicer for the reader of code in the future to have the name passed alongside the boolean value.

LouisFettet:
@hinzed1127 @Benaiah I'll need y'all's eyes on my latest changes. It simplifies the URL structure (so instead of including our index name in the URL we just have a simple `q` parameter and `page` parameter) and fixes [the routing bug I described in Slack](https://adhoc.slack.com/archives/C04EYUA76BT/p1691529754494399).

An unfortunate side effect of the navigation fix is that it locks us into a much more rigid type definition for setting the URL, so for a short-term fix I disabled TS on those lines. Before merging we should either:
* agree that disabling TS is okay here for a simpler URL (assuming we can't find a quick fix to the issue)
* go back to the more complex URL so that we're not disabling TS

There's also an issue where we can out put `undefined` for a search result if the `metaTitle` and `metaDescription` aren't set on the `Page details` entry (try searching for `bees` or `news`; Algolia gets a match on the `slug` but they don't have other information). Hiding these results in the template isn't a good solution since the results counter will be inaccurate, but I'm not sure what will be the easiest way to filter these results out. Perhaps we remove them from the `pageMetadataMap` for search only? Alternatively, we could see how UX folks feel about giving each `Page details` entry a title and description before Friday, which would at least give us a short-term fix.

LouisFettet:
@Benaiah to look into getting TS to cooperate in a follow-up. Merging!

PR #358, opened 2023-08-05
Don't upscale content images

Benaiah:
## Proposed changes

- Adds `max-width` and `max-height` to the image container if the image size type is not full bleed.
- Makes "static" on option for the "sizeType" of an image, which sets its `sizes` attribute to its static size (but _not_ its srcset attribute, so multiple DPI options

## Screenshots

image

- [x] Manually tested with QR code

### DIFF
--- a/src/lib/components/Header/Header.svelte
+++ b/src/lib/components/Header/Header.svelte
@@ -46,6 +46,7 @@
     alt=""
     width={headerBgWidth}
     height={headerBgHeight}
+    neverUpscaleImage={false}
     fit
   />
   <div class="ldaf-nav usa-navbar">
--- a/src/lib/components/Header/Header.svelte
+++ b/src/lib/components/Header/Header.svelte
@@ -46,7 +46,7 @@
     alt=""
     width={headerBgWidth}
     height={headerBgHeight}
-    neverUpscaleImage={false}
+    canUpscaleImage={true}
     fit
   />
   <div class="ldaf-nav usa-navbar">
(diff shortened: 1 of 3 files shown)


LouisFettet:
Noticing a couple issues during manual click test:
* The header background image doesn't stretch properly on my large monitor (2560px width window). We might want to override this rule for this image, but I know they also want to replace that image (see [LDAF-357](https://ldaf.atlassian.net/browse/LDAF-357)). We shouldn't merge this as-is though since it appears rather broken.
* I think the image under the "Louisiana Grown" accordion on [/food/local](https://ldaf-git-benaiah-dont-upscale-content-images-ldaf.vercel.app/food/local) should probably be full-width.  We should let our UX folks know this so the y can replace it in Contentful.

src/lib/components/Header/Header.svelte
```diff
@@ -46,6 +46,7 @@
     alt=""
     width={headerBgWidth}
     height={headerBgHeight}
+    neverUpscaleImage={false}
```
LouisFettet, line 49: Small nit: I'd prefer if we could avoid a double negative here; maybe `upscaleImage` instead? Not blocking if you'd rather leave it as is.

LouisFettet (approved):
LGTM, I let the UX folks know about the smaller image on the "Louisiana Grown" service in Slack, so feel free to merge!

src/lib/components/Header/Header.svelte
Benaiah: How about `canUpscaleImage`? `upscaleImage={true}` would incorrectly indicate that the image is _always_ upscaled.

PR #362, opened 2023-08-08
Remove unused pages

Benaiah:
## Proposed changes

- Removes `/documentation`, `/news`, and `/test-contentful-content`.


LouisFettet (approved):
LGTM, only suggestion woudl be to maybe change this line to a page that we're guaranteed to build every time, like `/.build-info`? https://github.com/la-ldaf/ldaf-site/blob/87e521706dee6a1ebd69246cbedf907104bc901a/src/routes/__tests__/homePageTestContent.ts#L29

PR #364, opened 2023-08-08
Implement events page

Benaiah:
Jira ticket: [LDAF-363](https://ldaf.atlassian.net/browse/LDAF-363)

## Proposed changes

- Implements `/about/events` and `/about/events/[page]` endpoints, both of which use the same loading and display logic.

## Screenshots

image

- [x] Manually tested
- [x] Tests continue to pass

[LDAF-363]: https://ldaf.atlassian.net/browse/LDAF-363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/routes/(infoPages)/about/events/event/[dateAndSlug]/+page.server.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+`;
+
+export const load = (async ({ params: { dateAndSlug }, parent }) => {
+  // dateAndSlug should be constructed like 2023-08-10-some-slug
+  const [_, dateString, slug] = dateAndSlug.match(/(\d{4}-\d{2}-\d{2})-([a-z1-9-]+)/) ?? [];
+  if (!dateString || !slug) throw error(404);
+  const date = new Date(dateString);
+  const dateStart = date.toISOString();
+  const dateEnd = new Date(date.getTime() + 1 * day).toISOString();
```
Comment (LouisFettet) [inline src/routes/(infoPages)/about/events/event/[dateAndSlug]/+page.server.ts:69]: Not important for right now, and I haven't looked into this much myself... I know we're defaulting to central time (`UTC-05:00`) in Contentful, but is there anything we need to do in the rare case that a different timezone is used? Do we know if daylight savings time will be accounted for?

src/routes/(infoPages)/about/events/event/[dateAndSlug]/+page.server.ts
```diff
@@ -0,0 +1,98 @@
...
+    baseBreadcrumbsPromise,
+    eventDataPromise,
+  ]);
+  const [event] = eventData?.eventEntryCollection?.items ?? [];
+  if (!event) throw error(404);
+  return {
+    event,
+    breadcrumbs: [
+      ...baseBreadcrumbs,
+      { title: event.shortTitle, link: `/about/events/event/${dateAndSlug}` },
```
Comment (LouisFettet) [inline src/routes/(infoPages)/about/events/event/[dateAndSlug]/+page.server.ts:95]: We should include a "fake" `pageMetadata` here to at least provide a `metaTitle` matching the event title and `url` so that we set the appropriate tags in the `<head>`.

src/routes/(infoPages)/about/events/shared.server.ts
```diff
@@ -0,0 +1,110 @@
...
+import { error } from "@sveltejs/kit";
+import type { PageServerLoad } from "./page/[page]/$types";
+import type { Breadcrumbs } from "$lib/components/Breadcrumbs";
+import { events as testEvents, pages as testEventPages } from "./__tests__/eventsTestContent";
+
+const limit = 20;
+
+export const query = gql`
+  query Events($limit: Int = 20, $skip: Int = 0) {
+    eventEntryCollection(limit: $limit, skip: $skip, order: [eventDateAndTime_DESC]) {
```
LouisFettet, line 15: Not sure if this is the appropriate place, but we'll want to improve this query so that we're filtering out past events and sorting in ascending order.

src/routes/(infoPages)/about/events/shared.server.ts
```diff
@@ -0,0 +1,110 @@
...
+    ) {
+      break fetchData;
+    }
+    return {
+      currentPageNumber: pageNumber,
+      totalPages: Math.ceil(eventsData.eventEntryCollection.total / limit),
+      events: eventsData.eventEntryCollection.items,
+      totalEvents: eventsData.eventEntryCollection.total,
+      breadcrumbs: breadcrumbsPromise,
+    };
```
LouisFettet, line 107: Same idea here with "fake" `pageMetadata` so that we're setting the page title. Maybe like "Upcoming Events, Page `pageNumber`"

LouisFettet (approved):
I'd like to get this through, so approving with some stipulations. Most of my comments can wait until later, so I'd suggest adding `//TODO` comments for anything you can't knock out quickly.

I think the only thing that really needs to happen before we merge is the filtering out of past events and sorting the upcoming events.

PR #365, opened 2023-08-09
Remove unimplemented pages for Offices and for Boards and Commissions.

LouisFettet:
Very straightforward since most changes happened in Contentful. Here we're just removing the hard-coded routes for the pages we don't have implemented yet ahead of launch:
* office aggregation page (`/about/organization`)
* boards and commissions aggregation page (`/about/boards-commissions`)
* individual board or commission pages (`/about/boards-commissions/[board-or-commission]`)

See [discussion in Slack](https://adhoc.slack.com/archives/C04EL8QK5FU/p1691606809213569) for additional context.

If y'all know a way we can archive these routes instead, that might be more ideal so that we don't have to re-add them in later. This seemed to be the quickest fix though.


Benaiah (approved):
We should probably remove the menu entries in Contentful before merging this PR.

LouisFettet:
> We should probably remove the menu entries in Contentful before merging this PR.

[Discussed in Slack](https://adhoc.slack.com/archives/C04EYUA76BT/p1691612361423239?thread_ts=1691611974.076629&cid=C04EYUA76BT), we're all good here. Merging!

PR #367, opened 2023-08-09
LDAF-265 Updates for launch (robots.txt and sitemap.xml)

LouisFettet:
Jira ticket: [LDAF-265](https://ldaf.atlassian.net/browse/LDAF-265)

**Do not merge until we confirm traffic to `www.ldaf.la.gov` is routing through Vercel.**

## Proposed changes

- Adds `/sitemap.xml` generated based on the `pageMetadataMap` (updated with changes from #357 so that we skip building breadcrumbs).
- Updates `robots.txt` to allow crawling all routes except `/.build-info`

## Requested feedback

See self-review, and let me know if you can think of anything else that needs to happen on Monday morning!

[LDAF-265]: https://ldaf.atlassian.net/browse/LDAF-265?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/routes/sitemap.xml/+server.ts
```diff
@@ -0,0 +1,32 @@
...
+    .map(([path, _]) => `<url><loc>https://ldaf.la.gov${path}</loc></url>`)
+    .sort()
+    .join("");
+  // TODO: Include tags other than just <loc> within <url>, like:
+  //       * <lastmod> for date of last Contentful update
+  //       * <changefreq> for likeliness of page to change since last crawl;
+  //         we should probably default to 'weekly' but set some pages like
+  //         past events to 'never'
+  //       * <priority> for highlighting important pages; will require
+  //         updates to the content model
```
LouisFettet, line 16: Let me know if y'all think we should go ahead and add any of these now; I could see marking the `<changefreq>` as `"weekly"` for all URLs being a safe default.

static/robots.txt
```diff
@@ -1,2 +1,4 @@
 User-agent: *
-Disallow: /
\ No newline at end of file
+Disallow: /.build-info
```
LouisFettet, line 2: Based on [this Google Devs doc](https://developers.google.com/search/docs/crawling-indexing/robots/robots_txt#url-matching-based-on-path-values), I think we could also do this:
```suggestion
Disallow: /.
```
if we wanted to ignore any path under the root that begins with `.`.  The period doesn't need to be escaped since it doesn't have a special meaning to the parsers like it does with regular expressions, and we don't need a wildcard like `/.*` since trailing wildcards are ignored (see `/fish*` example in the docs).

LouisFettet (commented):
self-review

PR #369, opened 2023-08-10
Update URLs for generating Lighthouse reports.

LouisFettet:
Noticed that our Lighthouse report index page was running into an error due to the removal of the `/documentation` page, and I figured this would be a good time to throw a Top Tier page and Core Content page into the mix.

### DIFF
--- a/.lighthouse/urls.txt
+++ b/.lighthouse/urls.txt
@@ -1,2 +1,3 @@
-http://localhost:4173/
-http://localhost:4173/documentation
+http://localhost:4173
+http://localhost:4173/land
+http://localhost:4173/land/indian-creek


.github/workflows/tests_and_reports.yml
```diff
@@ -148,17 +158,19 @@ jobs:
...
 
       - name: Generate Lighthouse reports
         run: |
-          npm run build
-          npm run preview &
-          sleep 5
-          lighthouse-batch --use-global --html --out lighthouse-reports/mobile --file .lighthouse/urls.txt --params "--only-categories=\"performance,accessibility,best-practices,seo\""
-          lighthouse-batch --use-global --html --out lighthouse-reports/desktop --file .lighthouse/urls.txt --params "--only-categories=\"performance,accessibility,best-practices,seo\" --preset=desktop"
+          lighthouse-batch --use-global --html --out lighthouse-reports/mobile --file .lighthouse/urls.txt --params "--only-categories=\"performance,accessibility,best-practices,seo\" --throttling.cpuSlowdownMultiplier=8"
+          lighthouse-batch --use-global --html --out lighthouse-reports/desktop --file .lighthouse/urls.txt --params "--only-categories=\"performance,accessibility,best-practices,seo\" --throttling.cpuSlowdownMultiplier=8 --preset=desktop"
```
Benaiah, line 173: Do these throttling settings come from somewhere specific? Are these to more closely match user devices? I was under the impression that GitHub Actions runners are already pretty slow.

Benaiah (approved):
Looks great! Looks like that youtube embed is as painful for our performance as I was afraid of. Just have one small question on the CPU throttling settings.

.github/workflows/tests_and_reports.yml
LouisFettet: This was suggested in the initial set of reports generated, since the GitHub Actions runner was struggling to keep up with the base slowdown modifier of `4`. I'll add a comment to the workflow to note this.

PR #372, opened 2023-08-11
Reimplement youtube subscribe button, avoiding loading google js

Benaiah:
## Proposed changes

Uses a styled link resembling the YouTube subscribe button to replace the embedded subscribe button and avoid all the extra JS it was loading. This also fixes the errors that were happening when navigating around the site if a `VideoCard` had _ever_ been loaded.

## Screenshots

image

- [x] Tested manually
- [x] Tests continue to pass


LouisFettet (approved):
Thanks for this, should improve our performance _way more_ than it reasonably should 😅 

Just have a couple minor improvement suggestions, but after that this should be good to go!

PR #375, opened 2023-08-11
Filter out the home page and null URLs from search.

LouisFettet:
Discussed during our "think tank" meeting, we need to make sure we're not linking to pages that don't exist, and there's not much use in linking to the home page.

### DIFF
--- a/src/routes/api/v1/algolia/+server.ts
+++ b/src/routes/api/v1/algolia/+server.ts
@@ -4,5 +4,6 @@ import { loadPageMetadataMap } from "$lib/loadPageMetadataMap";
 
 export const GET = (async () => {
   const { pageMetadataMap } = await loadPageMetadataMap({ includeBreadcrumbs: false });
-  return json(Array.from(pageMetadataMap.values()));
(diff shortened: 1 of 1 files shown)


hinzed1127 (approved):

src/routes/api/v1/algolia/+server.ts
```diff
@@ -4,5 +4,6 @@ import { loadPageMetadataMap } from "$lib/loadPageMetadataMap";
 
 export const GET = (async () => {
   const { pageMetadataMap } = await loadPageMetadataMap({ includeBreadcrumbs: false });
-  return json(Array.from(pageMetadataMap.values()));
+  // exclude the home page and pages where we can't determine the URL
+  return json(Array.from(pageMetadataMap.values()).filter((page) => page.url && !page.isRoot));
```
hinzed1127, line 8: note that right now, this in isolation doesn't filter out these results until we update the Algolia index directly. Uploading this entire response as a fresh set of records into Algolia _would_ remove them.

I'm working on a POST method to compliment this that would update things as needed

PR #379, opened 2023-08-11
Make isDocument check reactive

Benaiah:
## Proposed changes

If the prop it's checking is reactive, so should the check


LouisFettet (commented):
This seems like a safe check, but can you explain what would happen without it?

Benaiah:
> This seems like a safe check, but can you explain what would happen without it?

Right now the check is only run the _first_ time that the component renders, so if the component re-renders with different props the check won't run. It's not very important it's just a tiny consistency thing.

LouisFettet (approved):

PR #380, opened 2023-08-14
Stop leaking a redis client connection descriptor every request

Benaiah:
## Proposed changes

- Cache the KV client to avoid creating and connecting a new Redis client every request, which leaks file descriptors.

It's rather difficult to test if this is actually working or not, either manually or automatically. Existing tests pass, and I'm working on verifying that we're not leaking FDs.

Update: I was able to reproduce the problem locally and verify the fix by checking the open file descriptor list in Activity Monitor.

- [x] Tests continue to pass
- [x] Manually verified that site works
- [x] Manually verified that open file descriptors do not increase on every request

### DIFF
--- a/src/lib/services/server/kv/index.ts
+++ b/src/lib/services/server/kv/index.ts
@@ -1,4 +1,4 @@
-import { createClient as createRedisClient, type RedisClientType } from "redis";
+import { createClient as createRedisClient } from "redis";
 
 export type Client = {
   getBlurhashByURL: (url: string) => Promise<string | null>;
--- a/src/lib/services/server/kv/index.ts
+++ b/src/lib/services/server/kv/index.ts
@@ -1,4 +1,4 @@
-import { createClient as createRedisClient } from "redis";
+import { createClient as createRedisClient, type RedisClientType } from "redis";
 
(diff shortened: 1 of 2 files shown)


src/lib/services/server/kv/index.ts
```diff
@@ -11,37 +11,78 @@ const keys = {
...
+
+export const getClient = async ({ url = "redis://localhost" }: ClientOptionsInit = {}) => {
+  const clientOptions = { url };
+  const key = getClientKey(clientOptions);
+  const existingClient = clients.get(key);
+  if (existingClient) return existingClient;
+  const newClient = await createClient(clientOptions);
+  clients.set(key, newClient);
+  return newClient;
+};
```
LouisFettet, line 88: Apologies if you're still working on this, but are we calling this anywhere yet?

Benaiah: Sorry, I was just testing that my change actually prevented the open file descriptor issue. It seems to do so!

LouisFettet (approved):
Code changes LGTM; if tests pass and everything looked good in your testing then this should be good to go!

PR #383, opened 2023-08-17
Update dependencies, run formatter, migrate to Svelte 4

Benaiah:
Replaces #381, #382, #346, #331, #336, #370, #353, #337.

## Proposed changes

- Updates a bunch of dependencies
- Runs `npm run format` to apply new Prettier defaults
- Migrates to Svelte 4, which produces smaller and faster bundles

- [x] Lint passes
- [x] Check passes
- [x] Tests pass
- [x] Manually tested

## Other details

The easiest way to review this PR will be to go commit-by-commit, that way you can ignore the prettier diff. Each commit is a single small change.

You can also review the [migration checklist](https://github.com/la-ldaf/ldaf-site/blob/benaiah/update-some-dependencies/svelte-4-migration-checklist.md) which is based on the Svelte 4 upgrade guide.


src/lib/components/ContentfulRichText/nodes/Hr.svelte
```diff
@@ -1,14 +1 @@
...
-
-  export let node: NodeType;
-
-  let hr: Hr;
-  if (!isHr(node)) {
-    throw new Error("node is not an hr");
-  }
-  hr = node;
-</script>
-
```
LouisFettet, line 13: Was seeing some warnings popping up after this was removed; I think we can do this instead?

```
<script lang="ts">
  import type { Node as NodeType } from "@contentful/rich-text-types";
  import { isHr } from "../predicates";
  export let node: NodeType;
  if (!isHr(node)) {
    throw new Error("node is not an hr");
  }
</script>
```

svelte-4-migration-checklist.md
LouisFettet, line 1: Just remember to remove this before merge (or move it into docs).

LouisFettet (commented):
Thanks for taking this on! Overall this is a great set of changes but I am worried about some of the dependencies.

package.json
```diff
@@ -18,88 +18,89 @@
     "precheck": "./scripts/precheck.ts && npm run gen-operations",
     "check": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json --compiler-warnings \"css-unused-selector:ignore\"",
     "check:watch": "npm run check -- --watch",
-    "lint": "prettier --plugin-search-dir . --check . && eslint .",
-    "format": "prettier --plugin-search-dir . --write .",
+    "lint": "prettier --plugin=prettier-plugin-svelte --check . && eslint .",
+    "format": "prettier --plugin=prettier-plugin-svelte --write .",
```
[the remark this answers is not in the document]
Benaiah: Yeah, it doesn't run prettier on the `.svelte` files without explicitly including the plugin on the command line.

package.json
```diff
@@ -18,88 +18,89 @@
...
-    "tslib": "^2.6.0",
+    "tslib": "^2.6.1",
     "typescript": "^5.1.6",
-    "vite": "^4.4.8",
+    "vite": "^4.4.9",
     "vite-bundle-visualizer": "^0.8.1",
     "vite-imagetools": "^5.0.5",
-    "vite-plugin-svelte-purgecss": "^0.0.3",
-    "vitest": "^0.33.0"
+    "vite-plugin-svelte-purgecss": "^0.2.7",
```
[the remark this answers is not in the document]
Benaiah: Sure, I'll remove this dependency update from this PR.

LouisFettet:
Looks like the Announcement Banner icon still isn't rendering in Storybook (looks fine if I add it to the home page). We should try to track down which dependency is breaking that (I would assume it's Storybook) and either make a fix or save that upgrade for a separate PR as well.

Benaiah:
I had a really hard time trying to figure out what the problem was because it only happens in Storybook when you build it statically, not in Storybook dev mode. The issue is something to do with the `src` attribute of the `img` not being properly escaped, so the `src` attribute ends early (at the first double quote), everything gets malformed, and the DOM gets completely messed up; but only in Storybook and only when building statically. The problem also goes away entirely if we switch the `img` out for an `Icon`. But the `Icon` is just rendering an `img` with the same `src` under the hood so I don't know why that fixes it. Personally I'd prefer to keep the Storybook upgrade in place so that we can stop using an alpha release, but I understand if you'd prefer to try rolling back Storybook instead of using the workaround.

vite-plugin-ldaf-icon/index.ts
```diff
@@ -10,7 +10,7 @@ const vitePluginLDAFIcon = async () => {
     const iconPathWithinImg = iconPath.replace(/^node_modules\/@uswds\/uswds\/dist\/img\//, "");
     return [path.basename(iconPath).split(".")[0].replace(/^.+\//, ""), iconPathWithinImg];
   });
-  const iconNames = iconNamesAndPathsWithinImgDir.map(([name]) => name);
+  const iconNames = iconNamesAndPathsWithinImgDir.map(([name]) => name).sort();
```
LouisFettet, line 13: Does the sort order matter?

LouisFettet (approved):
LGTM, just have a minor question. As long as everything passes after pulling in the latest updates from `main` this should be good to merge!

vite-plugin-ldaf-icon/index.ts
Benaiah: The only reason for this was to make it easier to find icon names in the Storybook stories for the Icon (which is the only place that imports the list of icon names). We could move the sorting to Storybook if you think it'd fit better there.

LouisFettet: Oh okay that makes sense. No changes needed, I was just curious.

PR #394, opened 2023-08-17
Move YouTube API interactions to server, unload embed until user interaction

Benaiah:
## Proposed changes

- Loads the snippet data of YouTube videos from the server instead of the client
- Instead of immediately loading the YouTube embed, loads its thumbnail using the `Image` element and the thumbnails from the snippet
- On click or keyboard select, loads the YouTube embed, which will then autoplay if autoplay is supported and turned on

## Screenshots

image

- [x] Checks and tests pass
- [x] Manually tested

### Alternate solutions

We could continue to load the YouTube embed as soon as the user loads the page, with a strong negative performance impact.

### Possible drawbacks

On certain devices that don't allow videos to autoplay, the embed may require two clicks, taps, or keyboard interactions in order to play the video.


src/lib/components/VideoCard/VideoCard.svelte
```diff
@@ -1,78 +1,151 @@
...
+          allowfullscreen
+        />
+      </noscript>
+      {#if showYoutubeEmbed}
+        <iframe
+          class="ldaf-video-embed"
+          src={`https://www.youtube-nocookie.com/embed/${youtubeVideoID}?rel=0&autoplay=1`}
+          title="Embedded YouTube video player"
+          frameborder="0"
+          allow="encrypted-media;"
```
LouisFettet, line 106: In order to get autoplay working, you'll also need to edit this line:
```suggestion
          allow="autoplay;encrypted-media;"
```

src/lib/components/VideoCard/VideoCard.svelte
```diff
@@ -1,78 +1,151 @@
...
   export let url = "";
-  export let customTitle: $$Props["customTitle"] = null;
-  export let customDescription: $$Props["customDescription"] = null;
+  export let title: $$Props["title"] = null;
+  export let description: $$Props["description"] = null;
   export let variation = "primary";
 
-  let youtubeVideoId: string | null;
+  const maxDescriptionLength = 500;
+  $: trimmedDescription = description?.slice(0, maxDescriptionLength);
```
LouisFettet, line 29: I'm fine with setting a maximum length, but we should still cut the description off on newlines to avoid some of the issues I was seeing during development (e.g. showing hashtags that don't do anything, URLs going too long and squishing the video, etc.).

LouisFettet (changes requested):
Excellent work, I think this is going to make for a much faster UX. I do have a handful of mostly minor comments below that we should address before merging though.

Also (didn't know where to put this note), with these changes we should be good to remove `PUBLIC_YOUTUBE_API_KEY` from `.env.example` and from `src/env.d.ts`. We can remove it from Vercel after merging.

LouisFettet:
Forgot to mention, those Lighthouse scores are looking great!

Comparing:
`main`: https://ldaf-meta.vercel.app/lighthouse/
to this branch: https://meta-site-394-ldaf.vercel.app/lighthouse/

src/lib/services/server/kv/index.ts
```diff
@@ -50,31 +54,44 @@ export const createClient = async ({
...
-        if (result !== "OK") {
-          // TODO: better logging
-          console.error(new Error("could not set blurhash in KV store"));
-        }
-      } catch (err) {
-        // TODO: better logging
-        console.error(err);
-      }
+      const result = await redisClient.set(`${keys.blurhashByURL}:${url}`, blurhash);
+      if (result !== "OK") throw new Error("could not set blurhash in KV store");
```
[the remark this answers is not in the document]
Benaiah: This is already being caught and logged - see the definition of `getMethod` which wraps each KV client method in a try/catch and tries to reconnect to Redis if it's disconnected first.

src/lib/components/VideoCard/VideoCard.svelte
```diff
@@ -1,78 +1,151 @@
...
   export let url = "";
-  export let customTitle: $$Props["customTitle"] = null;
-  export let customDescription: $$Props["customDescription"] = null;
+  export let title: $$Props["title"] = null;
+  export let description: $$Props["description"] = null;
   export let variation = "primary";
 
-  let youtubeVideoId: string | null;
+  const maxDescriptionLength = 500;
+  $: trimmedDescription = description?.slice(0, maxDescriptionLength).replace(/\s[^\s]*?$/, "");
```
LouisFettet, line 29: Instead of stripping out newlines, I think we should remove everything after the first newline and just return the first paragraph. That way we can usually render full sentences, without the need to overflow part-way. The 500 character cut-off is fine but FWIW we accept a maximum of 256 characters for the `videoSubhead` field [in Contentful](https://app.contentful.com/spaces/pc5e1rlgfrov/content_types/videoWrapper/fields).

Benaiah: This isn't actually stripping newlines, it's stripping "everything after and including the last whitespace" so that we don't cut things off in the middle of a word. I think we should keep this logic in case the first paragraph is too long, but I'll add your suggestion to cut it off at the first newline if that's shorter.

src/lib/components/VideoCard/VideoCard.svelte
```diff
@@ -1,78 +1,151 @@
...
-        allow="encrypted-media;"
-        allowfullscreen
-      />
+      <noscript>
+        <iframe
+          class="ldaf-video-embed"
+          src={`https://www.youtube-nocookie.com/embed/${youtubeVideoID}?rel=0`}
+          title="Embedded YouTube video player"
+          frameborder="0"
+          allow="autoplay;encrypted-media;"
```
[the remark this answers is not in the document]
Benaiah: I don't want the `noscript` version to autoplay, this was a mistake lol. Fixing now.

src/lib/components/VideoCard/VideoCard.svelte
```diff
@@ -1,78 +1,154 @@
...
+  export let title: $$Props["title"] = null;
+  export let description: $$Props["description"] = null;
   export let variation = "primary";
 
-  let youtubeVideoId: string | null;
+  const maxDescriptionLength = 500;
+  $: trimmedDescription = description
+    ?.replace(/\n.+$/, "\n")
+    .slice(0, maxDescriptionLength)
+    .replace(/\s[^\s]*?$/, "");
```
LouisFettet, line 32: Tested with a screen reader and the entire description gets read, regardless of how much is actually showing on the screen. On the Animals top tier page the description gets cut-off mid-sentence, so we should probably try to break on the first paragraph since that would be easier than trying to determine where one sentence ends and another begins.

LouisFettet (commented):
Sorry for all the back-and-forth, but I think we're getting close!

src/lib/components/VideoCard/VideoCard.svelte
Benaiah: This code was _supposed_ to break at the first paragraph _or_ 500 characters (while preventing mid-word breaks), whichever was smaller. Unfortunately I messed up the regex a little bit. I've pushed a fix.

src/lib/components/VideoCard/VideoCard.svelte
```diff
@@ -1,78 +1,154 @@
...
+  export let title: $$Props["title"] = null;
+  export let description: $$Props["description"] = null;
   export let variation = "primary";
 
-  let youtubeVideoId: string | null;
+  const maxDescriptionLength = 500;
+  $: trimmedFirstParagraph = description
+    ?.replace(/\n[\s\S]*$/, " ") // remove everything after the first newline
+    .slice(0, maxDescriptionLength) // limit to maxDescriptionLength
+    .replace(/\s[^\s]*?$/, ""); // remove everything after the start of the last whitespace character
```
LouisFettet, line 32: This is removing the last word on the following pages:
* [/plants](https://ldaf-git-benaiah-dont-load-youtube-embed-until-its-clicked-ldaf.vercel.app/plants)
* [/land](https://ldaf-git-benaiah-dont-load-youtube-embed-until-its-clicked-ldaf.vercel.app/land)
* [/business](https://ldaf-git-benaiah-dont-load-youtube-embed-until-its-clicked-ldaf.vercel.app/business)

Benaiah: good catch, should be fixed now

LouisFettet (approved):
LGTM! I think it would be nice to remove hashtags but that's probably a bit too much effort for something that editors can handle themselves by providing a custom description. :shipit: !

PR #398, opened 2023-08-18
Set meta tag with theme color

Benaiah:
## Proposed changes

Sets a meta tag with the theme color. This is the color used by browsers (primarily mobile browsers) to theme the address bar area. Matching it with the background color of the header should make it a smooth transition.

## Screenshots

image


LouisFettet (approved):
Looks good, maybe just add a comment noting the variable (`$primary-darkest`) so that we have a higher chance of catching this if we ever need to update the value.

PR #400, opened 2023-08-18
Resize top tier header images to avoid loading unshown parts

Benaiah:
## Proposed changes

- Makes top tier header images have a static maximum height
- Resizes top tier header images in `getSources` such that we never require more of an image than can possibly be shown

## Screenshots

Desktop:

image

Mobile:

image

### DIFF
--- a/src/routes/(infoPages)/layout.scss
+++ b/src/routes/(infoPages)/layout.scss
@@ -1,3 +1,4 @@
 .top-tier-page-hero-image {
-  height: 33vh;
+  height: 14rem;
+  max-height: 33vh;
 }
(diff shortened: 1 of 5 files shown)


src/routes/(infoPages)/layout.scss
```diff
@@ -1,3 +1,4 @@
 .top-tier-page-hero-image {
-  height: 33vh;
+  height: 14rem;
```
LouisFettet, line 2: This is definitely much cleaner, but the images can look overly-cropped at larger screen sizes. If we increase this value based on screen size (with static values, not `%`, `vh`, etc.) would that mess up any of the optimizations? I'm thinking maybe we increase it to `16rem` on `tablet` and then `18rem` on `desktop`?

Benaiah: Sure, I'll look into this!

PR #403, opened 2023-08-21
Add/Update server endpoints for page metadata map/Algolia updates, etc

hinzed1127:
Jira Ticket: [LDAF-425](https://ldaf.atlassian.net/browse/LDAF-425)

## Proposed changes

This adds and reorganizes a number of server endpoints, all under `/api/v1`:
  - `POST search-index`
      -  This is what [our Contentful webhook](https://app.contentful.com/spaces/pc5e1rlgfrov/settings/webhooks/07k8eHWY9xHpSQIUSS849x/settings) will point to, triggering subsequent calls to our Algolia index as needed.
  - `GET search-index/status` ([test site url](https://ldaf-ohb04sdo1-ldaf.vercel.app/api/v1/search-index/status))
        - This is something that started as a way I was debugging the other new endpoints, but I think it'll be useful to have as a way to see any differences that currently exist between records in Algolia and Contentful.
  - `POST search-index/reset`
        - This grabs the same array that `GET metadata-map` returns, and then adds the Algolia-required `objectID` to each item. This replaces each record with a matching objectID with the provided set of values.
  - `GET metadata-map` ([test site url](https://ldaf-ohb04sdo1-ldaf.vercel.app/api/v1/metadata-map))
        - This just renames/moves the current `/api/v1/algolia` endpoint that loads the metadata map and filters out null URLs and the root path.

Other implementation details:
  - I changed the `objectID` value in Algolia (what Algolia uses as a unique ID for index records) to match `sys.id` on Contentful content. This should make for easier updating of values within the endpoint functions.
 - Right now, we only care about content types of type 'pageMetadata'. In the future, this will expand to probably other things (e.g. serviceGroup, serviceEntry, and topTier content types) that will have useful information not in the metaTitle and metaDescription. The `POST search-index` endpoint is constructed so that we can make future updates to accommodate new content types relatively easily.
 -  Side note: Is there a way to do this with GraphQL or are we restricted to using `getContentTypes` on the Contentful JS client? This is currently hardcoded into the the update logic, but it would be nice to make it dynamic, if possible (I'm not sure  it is).

- [ ]  Webhook requests are successfully transformed by sveltekit server endpoints before calling and updating Algolia
- [ ]  Manually tested with postman calls to update Algolia, as well as using Webhook.site to validate proper webhook calls and to get accurate request bodies.

## Other details

- [This is the webhook in Contentful](https://app.contentful.com/spaces/pc5e1rlgfrov/settings/webhooks/07k8eHWY9xHpSQIUSS849x/settings). It's currently posting to a Webhook.site url for testing purposes, but would point to `/api/v1/search-index` once this is merged.
- [Algolia Search Index it will eventually post to](https://dashboard.algolia.com/apps/9IIDYROXZ5/explorer/browse/prod_ldaf?searchMode=search)
- [Test site that the webhook is currently posting to](https://webhook.site/#!/85db191c-e816-4750-98eb-5b3f00fd2a62/18005e48-5ed9-4f0c-b49b-9aa8d8f0f0ef/1). This is helpful in seeing the shape the request body from Contentful takes.
## Requested feedback

[LDAF-425]: https://ldaf.atlassian.net/browse/LDAF-425?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.eslintrc.cjs
```diff
@@ -46,6 +46,7 @@ module.exports = {
     "@typescript-eslint/no-unused-vars": [
       "warn",
       {
+        ignoreRestSiblings: true,
```
hinzed1127, line 49: This is accommodate filtering out values via destructuring like [below](https://github.com/la-ldaf/ldaf-site/pull/403/files#diff-948f3271a51b6b2180724f66411a30790e43053cd5f1236d8f9f53b821b1ed55R24)

hinzed1127 (commented):
I annotated some of the operations here more than usual, so self-review is a single comment, for now

src/routes/api/v1/search-index/reset/+server.ts
```diff
@@ -0,0 +1,50 @@
...
+import { ALGOLIA_API_KEY } from "$env/static/private";
+const algoliaClient = algoliasearch(PUBLIC_ALGOLIA_APP_ID, ALGOLIA_API_KEY);
+
+/**
+ * This endpoint does a wholesale reset of the page metadata map in Algolia.
+ * Since it requires each record to have an objectID, a reset call won't result
+ * in any duplication of metadata records in the Algolia index, just a replacement
+ * of the previous object value with the current value in contentful
+ * TODO: secure this somehow to prevent unnecessary operations
+ **/
```
hinzed1127, line 14: This isn't a security concern in the sense that it would wipe out the index, but for limiting overall operations in Algolia.

As it currently works, it just grabs the `pageMetdataMap` and updates all of those records in Algolia with the current values. Adding some kind of credentialed check in front of this call would be mainly to prevent anyone from racking up update calls, [which has monthly limits](https://support.algolia.com/hc/en-us/articles/4406981829777-How-does-Algolia-count-records-and-operations-).

src/routes/api/v1/search-index/status/+server.ts
```diff
@@ -0,0 +1,114 @@
...
+    expected_algolia_records: pageMetadataMap.size - nullURLs.length - missingInAlgolia.length - 1,
+    // non-zero values here indicate that new records aren't getting added properly
+    missing_in_algolia: missingInAlgolia,
+    // We should never encounter records in Algolia that aren't in Contentful.
+    // If we do, they've been unpublished/deleted and so should also be deleted from Algolia
+    missing_in_contentful: missingInContentful,
+    mismatched_record_total: mismatches.length,
+    nullURLs_in_contentful: nullURLs,
+    mismatches,
+  };
```
hinzed1127, line 111: TODO Idea: As long as the contentful webhook is functioning as intended, any mismatches between contentful and algolia _should_ remain zero.

However, it could be useful to have a cron job call this nightly, and if there are any mismatches or unexpected discrepancies, e.g., `results.mismatched_record_total !== 0`, then have the logger (once it's merged) post to slack

LouisFettet (commented):
This looks really good! I wonder if there's a way we can use `code-gen` to get some type definitions for the Contentful webhook request body?

Benaiah (changes requested):
This is definitely looking like the right direction! I think we need the following addressed before this gets merged though:

- There's no authentication for endpoints that trigger backend actions. Eventually we'll want to integrate this with the authentication we use for the previews, but for now we can just generate a static token, store it in an environment variable, and require it be set as a bearer token in the `Authorization` header.

- There are no types specified for the data that comes from / enters Algolia. We should either generate types from Algolia or set them up by hand. Right now there's a lot of `any` types.

It might be worth setting up an Algolia service in `lib/services/server` that wraps the regular Algolia client to handle returning and requiring more specific types. We could keep the types for the Algolia data in that folder as well.

src/routes/api/v1/search-index/reset/+server.ts
```diff
@@ -0,0 +1,53 @@
...
+     */
+    return {
+      ...metadataValue,
+      objectID,
+    };
+  });
+
+  try {
+    // TODO: investigate changing to `index.replaceAllObjects` for a "hard" reset
+    const response = await index.saveObjects(algoliaRecords);
```
LouisFettet, line 47: If I understand correctly, as it stands now, this operation will add and update records in Algolia but not remove any, right? If that's the case we might want to document that here or in the comment you already have above `POST`.

LouisFettet (commented):
Looking really good! Only thing I think that needs to change is how we're handling the environment variables; hopefully that's an easy fix.

I didn't manually test anything locally besides the `status` endpoint, but the code for the other endpoints makes sense to me, and I assume you've already performed your own manual testing a half a hundred times by now 😅 

I think it would be really useful to have some unit tests set up here, but that can be handled in a follow-up so we can get this out the door.

.github/workflows/tests_and_reports.yml
```diff
@@ -30,6 +30,9 @@ jobs:
   lint-and-unit-tests:
     name: "ESLint, TypeScript Check, and Unit Tests"
     runs-on: ubuntu-latest
+    env:
+      ALGOLIA_API_KEY: ${{ secrets.ALGOLIA_API_KEY }}
+      AUTH_TOKEN: ${{ secrets.AUTH_TOKEN }}
```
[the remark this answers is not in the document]
hinzed1127: I think this is at least the third time I've gotten turned around with the environment variables like this 🤦‍♂️ . Updated to your suggestions and everything appears to be working now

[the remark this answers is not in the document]
LouisFettet: Good to hear! Wasn't sure if there was something else stopping us from doing it this way. When you get a sec can you remove the new env vars from the [Actions secrets](https://github.com/la-ldaf/ldaf-site/settings/secrets/actions)?

src/routes/api/v1/search-index/reset/+server.ts
hinzed1127: Yeah that's a good callout. Adding some more comments to make that more explicit

.github/workflows/tests_and_reports.yml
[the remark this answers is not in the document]
hinzed1127: good catch. done!

LouisFettet (approved):
:shipit: (you'll have to use admin priveleges to override Benaiah's request for changes)

PR #407, opened 2023-08-23
Fix Vercel Speed Insights integration

Benaiah:
I thought that we would be able to name the analytics ID environment variable whatever we wanted but it seems to be hardcoded as `VERCEL_ANALYTICS_ID`. This PR should fix things.


LouisFettet (approved):
LGTM, maybe just add a comment that we're treating the env var as public and this should be good to go!

PR #412, opened 2023-08-24
Only send speed insights in production; load `web-vitals` on demand

Benaiah:
We're currently sending data to Speed Insights even in development and preview settings.


LouisFettet (approved):
Makes sense to me!

PR #413, opened 2023-08-24
Implement "Popular resources" section on homepage

Benaiah:
Jira ticket: [LDAF-395](https://ldaf.atlassian.net/browse/LDAF-395)

## Proposed changes

- Implements "Popular resources" section on homepage

## Screenshots

image

- [x] Tested manually
- [x] Existing tests and checks pass

[LDAF-395]: https://ldaf.atlassian.net/browse/LDAF-395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/routes/+page.server.ts
+++ b/src/routes/+page.server.ts
@@ -12,6 +12,7 @@ import type { PageMetadataMapItem } from "$lib/loadPageMetadataMap";
 import type { ExtractQueryType } from "$lib/util/types";
 import homePageTestContent from "./__tests__/homePageTestContent";
 import type { YoutubeVideoData } from "$lib/services/server/youtube/getYoutubeVideoData";
+import type { ResourceLinks } from "$lib/components/ResourceLinks";
 
 const query = gql`
   query HomeCollection($metadataID: String!) {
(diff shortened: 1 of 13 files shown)


src/routes/+page.server.ts
```diff
@@ -29,31 +30,38 @@ const query = gql`
...
+        }
         featuredServiceCardsCollection {
           items {
-            ... on ServiceGroup {
-              pageMetadata {
-                ... on PageMetadata {
+            pageMetadata {
+              sys {
+                id
+              }
```
LouisFettet, line 49: I thought that `... on ServiceGroup`, `... on PageMetadata`, and other inline fragments were required whenever we were using a Contentful reference. Is that not the case?

src/routes/+page.svelte
```diff
@@ -87,4 +89,8 @@
       {/each}
     </ul>
   {/if}
+  {#if popularResources && popularResources.length > 0}
+    <h2 class="ldaf-homepage__popular-resources-heading font-sans">Popular resources</h2>
```
LouisFettet, line 93: Seems like `font-sans` isn't doing anything?

src/routes/page.scss
```diff
@@ -0,0 +1,8 @@
+@use "uswds-core" as * with (
+  $theme-show-notifications: false
+);
+
+.usa-prose > .ldaf-homepage__popular-resources-heading {
+  margin-bottom: 1rem;
+  @include u-font("sans", $theme-h2-font-size);
```
LouisFettet, line 7: Would you mind applying the "sans" font to the card titles while you're here? Just noticed they don't match the mocks (both the featured service cards on the home page and the other cards across the site look like they should be using a sans font heading).

I'm sure there are other places we need to update the fonts but this seems like a safe change to make as a one-off here.

LouisFettet (approved):
Looks great! Just have a handful of minor comments, suggestions, and questions. Nothing that should block merging.

src/routes/+page.server.ts
Benaiah: I'm pretty sure that's only necessary if there's more than one type that you're querying on, like there is in a heterogenous list. This works and goes thru `npm run check` so I'm in favor of keeping it.

PR #423, opened 2023-08-29
Fix stale accordion-expanded state when navigating between pages

Benaiah:
Jira ticket: [LDAF-371](https://ldaf.atlassian.net/browse/LDAF-371)

## Proposed changes

- Adds an `each` key that is the accordion item ID when rendering child accordion components. This fixes a bug where the state and event handlers were being inappropriately preserved after page navigation, leading to an accordion item that is open by default (instead of closed) and takes two clicks to close (because the event handler is stale).

- Slightly refactors `Accordion`.

Right now it's still not perfect because the expandedItems record is still sticking around after navigation, it's just not causing problems because we never use accordion items with the same ID on separate pages.

- [x] Tests and checks pass
- [x] Manually verified that bug is no longer present

[LDAF-371]: https://ldaf.atlassian.net/browse/LDAF-371?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/components/Accordion/Accordion.svelte
+++ b/src/lib/components/Accordion/Accordion.svelte
@@ -1,16 +1,16 @@
 <script lang="ts">
-  import { setContext } from "svelte";
   import { writable } from "svelte/store";
+  import { setAccordionContext } from "./context";
 
   export let multiselectable = false;
   export let bordered = false;
 
-  const selectedItems = writable<Record<string, boolean>>({});
+  const expandedItems = writable<Record<string, boolean>>({});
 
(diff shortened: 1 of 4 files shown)


LouisFettet (approved):
LGTM! This also resolves [LDAF-371](https://ldaf.atlassian.net/browse/LDAF-371) if you don't mind throwing that into the PR description.

PR #430, opened 2023-09-06
[LDAF-366] Add Google Analytics

LouisFettet:
Jira ticket: [LDAF-366](https://ldaf.atlassian.net/browse/LDAF-366)

## Proposed changes

- Adds an `<Analytics/>` component used to set up GA. The default setup will automatically send data for page views, form completions, etc. When not running in Vercel production, we set the `debug_mode` flag to `true` during setup.
- Adds our Google Analytics ID to public env vars.
- Adds dependencies for `partytown` (which manages a service worker thread to keep third-party scripts from blocking the main thread) and for GA type definitions (`@types/gtag.js`).
- Adds custom events for expanding / collapsing service accordions and for clicking service CTAs.

## GA Info

* Team members have been added to dashboard, which [you can view here](https://analytics.google.com/analytics/web/#/p401901511/reports/intelligenthome).
* You can test out the debugger by visiting the [DebugView](https://analytics.google.com/analytics/web/#/a281047106p401901511/admin/debugview/overview)

GA is receiving events for:
- [x] page views
- [x] service accordion expand / collapse
- [x] service CTA clicks

## Requested feedback

See self-review.

[LDAF-366]: https://ldaf.atlassian.net/browse/LDAF-366?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.svelte
```diff
@@ -30,11 +30,28 @@
...
+        serviceAccordionCTA.classList.add("usa-button");
+        serviceAccordionCTA.addEventListener("click", () => {
+          if (window.gtag) {
+            window.gtag("event", "ldaf_service_cta_click", {
+              event_category: "click",
+              event_label: serviceAccordionCTA.textContent,
+              ldaf_service_name: serviceAccordionHeadingText,
+              ldaf_service_cta_destination: serviceAccordionCTA.getAttribute("href"),
+            });
+          }
```
Comment (LouisFettet) [inline src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.svelte:51]: I was originally thinking about breaking out an analytics "send event" function into a separate service, but I think it's simple enough that we don't need to do this. Only real reason to break it out would be to add retry logic, which I don't think is necessary (at least not yet).

src/routes/+layout.svelte
```diff
@@ -64,6 +71,16 @@
 </script>
 
 <svelte:head>
+  <script>
+    // https://partytown.builder.io/sveltekit
```
LouisFettet, line 75: Highly recommend giving this a look as well as the other docs to get a better understanding of what's going on here.

* [Primer](https://partytown.builder.io/how-does-partytown-work)
* [How forwarding works](https://partytown.builder.io/forwarding-events)
* [SvelteKit Tutorial](https://partytown.builder.io/sveltekit)

Some differences between the tutorial and what I've done here are:
* We also forward `gtag` so that we can call that instead of `dataLayer.push`, since I think that `gtag` is more clear (many things can use `dataLayer`, but `gtag` is specific to GA). We edit the boilerplate GA setup script so that we declare `gtag` on the `window` so we can access it in the main thread.
* We're currently not reverse-proxying any scripts. It seems like the latest version of `partytown` handles this automatically. If we notice CORS errors once deployed to Vercel, we can use [this older version of the SvelteKit tutorial](https://monogram.io/blog/add-partytown-to-svelte) as a rough guide to setting up rewrites in Vercel to handle reverse-proxying.
* We're doing a little extra work to ensure we're not hard-coding the GA ID (instead using an env var), which makes the whole implementation a little less easy-to-read but more maintainable.

LouisFettet (commented):
self-review!

Benaiah (approved):
A couple superficial suggestions which you can merge in at https://github.com/la-ldaf/ldaf-site/pull/449 (you can hit the merge button to merge them into this branch, because that's the base of that PR), but this looks good once those are brought in.

PR #461, opened 2023-09-18
Add deep links to accordion, auto-expand on page render

hinzed1127:
Jira ticket: [LDAF-378](https://ldaf.atlassian.net/browse/LDAF-378)

## Proposed changes

- constructs url hash for accordions based on their title
- accordion auto-expands when navigating to aforementioned link

- [ ] 
- [ ]  jsdom has limitations in testing browser location, refreshes, etc. The better option if we desire to test this is adding an e2e test for a specific page. This is depending on potentially shifting content, but could be worth adding if it adds some testing security, even with the small "potential for maintenance" cost

### Possible drawbacks

- the href is currently constructed from the accordion title. A more durable option would the title to come directly from contentful

## Requested feedback

[LDAF-378]: https://ldaf.atlassian.net/browse/LDAF-378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


LouisFettet (commented):
Looks good, just a couple notes before we merge:
* We should run the design / content of the link by the UX team if we haven't already. I'm curious if we can get away without the text and just show the icon if we give it appropriate accessible text.
* Using the `$page` store in a component breaks the [Storybook component](https://storybook-461-ldaf.vercel.app/?path=/docs/components-accordion--docs); maybe we can provide this data to the component from the Core Content page (since I think that's the only place we're using accordions)?
* We should use [the existing `slugify` util](https://github.com/la-ldaf/ldaf-site/blob/main/src/lib/util/slugify.ts) rather than a one-off in this component. You'll want to make sure that doing so doesn't break anything here. Maybe adding a test to that util would be useful to throw a few different edge cases at it.

hinzed1127:
@LouisFettet Addressing those in order:

> We should run the design / content of the link by the UX team if we haven't already. I'm curious if we can get away without the text and just show the icon if we give it appropriate accessible text.

this matches the [current spec](https://www.figma.com/file/oGKbyCnCRRdNzLYbiags93/LDAF-Component-Library-USWDS-3.0.2?type=design&node-id=2195-3296&mode=design&t=4wdChcdN6ZLNwi4v-0). I think the main thought behind including text is the typical visitor to LDAF might not immediately associate just the icon with being a url. Also, so we're not copying the url upon clicking, this is more explicit. 

> Using the $page store in a component breaks the [Storybook component](https://storybook-461-ldaf.vercel.app/?path=/docs/components-accordion--docs); maybe we can provide this data to the component from the Core Content page (since I think that's the only place we're using accordions)?

It looks like it _should_ support mocking stores ([ref](https://github.com/storybookjs/storybook/blob/next/code/frameworks/sveltekit/README.md#supported-features)), but if that's not quickly resolved, I'm fine with that modification

> We should use [the existing slugify util](https://github.com/la-ldaf/ldaf-site/blob/main/src/lib/util/slugify.ts) rather than a one-off in this component. You'll want to make sure that doing so doesn't break anything here. Maybe adding a test to that util would be useful to throw a few different edge cases at it.

Good call. I'll make that update

LouisFettet:
@hinzed1127  I probably should have used an `ol` instead of a `ul` for my feedback 😅 

1. RE: Design, thanks for the link; I completely missed this. In this case I think we just want to make a few styling tweaks to get it closer to the mock:
   * decrease the font size one tick
   * use `$grayscale-70` for the color
   * have it appear on its own line rather than wrapping content around it; should have about 2 margin units (16px) above and below
2. RE: Storybook, that's new! Yeah if you have a sec to take a stab at mocking the page store in the Storybook component that would be great. If it doesn't work then I would say spend no more than like 10-15 minutes trying to move the page store out of the component; otherwise I don't think the juice is worth the squeeze and we can deal with the Storybook component being broken.

src/lib/components/Accordion/AccordionItem.svelte
```diff
@@ -11,9 +17,28 @@
...
+        .toLowerCase()
+        .split(" ")
+        .join("-")
+    : null;
+
+  onMount(() => {
+    if ($page.url.hash === `#${titleHref}` && $expandedItems) {
+      $expandedItems[id] = true;
+    }
+  });
```
Benaiah, line 38: Let's use `afterNavigate` instead of `onMount`. `afterNavigate` will run whenever `onMount` does, but _also_ when navigating to a new URL.

Benaiah (changes requested):
Small change request, but otherwise looks good to me once @LouisFettet's suggestions are also handled.

src/lib/util/slugify.ts
```diff
@@ -1,8 +1,10 @@
...
 export default (s: string) =>
   s
     .toLowerCase()
-    .replaceAll(/[^a-z\s-_]/g, "")
-    .trim()
-    .replaceAll(/\s+/g, " ")
-    .replaceAll(" ", "-");
+    .replaceAll(/[^a-z0-9\s-&_]/g, "")
+    .split(" ")
+    .join("-");
```
Benaiah, line 10: This change has some improvements (notably, number handling) but also has a couple issues:

- `&` should _not_ be used in slugs. It works in this case because the slug is only used as part of the hash, but if this function was used to generate a route it would break in surprising ways.
- This function removes the replacement of multiple consecutive whitespaces with a single space, which can lead to doubled hyphens in the slug.
- This function removes the `.trim` call, which can lead to hyphens at the beginning or end of the slug.
- This function splits on individual spaces, not sequences of whitespace, which can lead to empty strings in the slug leading to doubled hyphens or retaining whitespace that is not a space in the slug.

Something like this should work:

```ts
export default (s: string) =>
  s
    .toLowerCase()
    .replaceAll(/[^a-z0-9\s-_]/g, "")
    .trim()
    .replaceAll(/\s+/g, "-")
```

Benaiah (changes requested):
Looks good! A couple issues with the changes to `slugify` but they should be quick to resolve. (If @LouisFettet approves your fixes to these changes feel free to dismiss my review and merge since I'm going to be out for much of the afternoon).

src/lib/components/Accordion/AccordionItem.svelte
```diff
@@ -57,7 +49,7 @@
   </button>
 </h3>
 <div class="usa-accordion__content usa-prose" {id} hidden={!expanded}>
-  <Link class="accordion-deeplink" href={`#${titleHref}`}
+  <Link class="accordion-deeplink font font-body-xs" href={`#${titleHref}`}
```
LouisFettet, line 52: I don't believe the standalone `font` class here is doing anything, is it?

LouisFettet (commented):
Should be good to go after addressing a couple more things:
* I'll echo @Benaiah's point about the ampersand; better to just strip it out since it could cause problems.
* The link is currently taking up the full width of the container; you might want to wrap it in a div so that only the link text and icon are clickable.
* There's a class that I think might be a no-op that I made a note of below.

src/lib/util/slugify.ts
hinzed1127: I modified this to have an additional `replaceAll` preceding the current one with `.replaceAll(/&/g, "and")`. This will preserve legibility of links using ampersands ([here's a few in the wild under "Resources](https://www.ldaf.la.gov/land/fire/safety/statewide-burn-ban-issued-due-to-concerning-dry-conditions)) without messing up any routing

LouisFettet (approved):
LGTM, :shipit:

PR #462, opened 2023-09-18
[LDAF-402] Use Contentful tags to render related news and events on Top Tier pages

LouisFettet:
Jira ticket: Part of [LDAF-402](https://ldaf.atlassian.net/browse/LDAF-402)

## Proposed changes

Using the new `subject` tags tied to Top Tier pages (seen [here in Contentful](https://app.contentful.com/spaces/pc5e1rlgfrov/settings/tags)), we can now pull in the latest 5 news entries and the 5 next upcoming events for a given tag and render them on the appropriate Top Tier page.

- Adds an additional query to Top Tier pages for news and event entries with a tag based on the page's slug. Every Top Tier page has an accompanying tag, e.g. `/animals` = `subject-animals`, `/plants` = `subject-plants`, etc.
  - News entries are filtered to the five most recent tagged entries and sorted by most recent to least recent.
  - Event entries are filtered to the five most imminent tagged entries and sorted by most imminent to least imminent.
- Updates to the event and news styling, including the default styling on the `<Date>` and `<Tag>` components (this is mostly all color changes)
-Small update to the date utility to calculate the date six months ago from today's date (used to limit the amount of news entries we show).
- Swaps out the snippets on News from the first paragraph of body text to the subheading field. This has the added benefit of reducing query complexity and simplifying rendering.
- Adds new props to the News and Event components to control the heading level of the entry title.

## Requested feedback

No self-review since I feel like the changes are mostly straightforward, but please don't hesistate to ask questions if anything doesn't seem immediately intuitive. Happy to add comments and such!

[LDAF-402]: https://ldaf.atlassian.net/browse/LDAF-402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/routes/(infoPages)/[topTierPage]/+page.server.ts
```diff
@@ -150,12 +204,33 @@ export const load = async ({
...
+    );
+
+    const [heroImageBlurhash, featuredServices, youtubeVideoData, taggedData] = await Promise.all([
       heroImageBlurhashPromise,
       Promise.all(featuredServicesPromises).then((arr) => arr.flat()),
       youtubeVideoDataPromise,
+      taggedDataPromise,
     ]);
 
+    if (!taggedData) break fetchData;
```
Comment (Benaiah) [inline src/routes/(infoPages)/[topTierPage]/+page.server.ts:232]: Should we fail `fetchData` if the `taggedData` doesn't come thru, or should we gracefully degrade the page?

Benaiah (approved):

src/routes/(infoPages)/[topTierPage]/+page.server.ts
Comment (LouisFettet) [reply src/routes/(infoPages)/[topTierPage]/+page.server.ts:232]: Good question; my thinking was that if `taggedData` doesn't come through it means the request failed (not that there was no tagged data, since that should be represented by zero-element arrays for `newsCollection.items` and `eventEntryCollection.items`), which probably means there's a larger problem.  I suppose it makes sense to try to render what we can if only this request fails.  Will update!

Comment (LouisFettet) [reply src/routes/(infoPages)/[topTierPage]/+page.server.ts]: Updated in bea86b2 so that we still render the page even if this request fails. Thanks for the call-out on this. Going to go ahead and merge now.

PR #470, opened 2023-09-20
[LDAF-402] Render latest news and upcoming events on the Home Page

LouisFettet:
Jira ticket: [LDAF-402](https://ldaf.atlassian.net/browse/LDAF-402)

This is pretty similar to #462; we want to display news and events on the homepage but in a slightly more compact style.

## Proposed changes
- As part of the homepage query, also request the 4 most recent news entries and the next 4 upcoming events, and render them on the homepage between "Popular resources" and "From the commissioner" sections.
- Adjust lots of spacing on the homepage to more closely match mocks.
- Allow variations on News and Events components to render them slightly differently on the homepage.
- Make some minor adjustments to the styling of the Event component (color, size, month capitalization)

## Requested feedback

See self-review.

[LDAF-402]: https://ldaf.atlassian.net/browse/LDAF-402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/components/Date/Date.scss
+++ b/src/lib/components/Date/Date.scss
@@ -2,11 +2,17 @@
   display: flex;
   flex-direction: column;
   text-align: center;
-  width: 78px;
-  height: calc(39px * 2 + 2px);
   font-size: 18px;
   border: 1px solid $primary-light;
   flex-shrink: 0;
+  &--big {
+    width: 78px;
+    height: 78px;
+  }
+  &--small {
+    width: 64px;
+    height: 64px;
+  }
 }
 
 .event-date-month,
@@ -23,6 +29,7 @@
 .event-date-month {
   background-color: $primary-light;
   color: $white;
+  text-transform: uppercase;
 }
 
 .event-date-day {
--- a/src/routes/(infoPages)/about/news/NewsEntrySnippet.svelte
+++ b/src/routes/(infoPages)/about/news/NewsEntrySnippet.svelte
@@ -9,8 +9,11 @@
 
   import type { PageServerData } from "./$types";
 
+  type NewsSnippetVariation = "default" | "homepage-featured" | "homepage-listed";
+
   export let entry: NonNullable<PageServerData["newsEntries"][number]>;
   export let headingLevel: HeadingLevel = 2;
+  export let variation: NewsSnippetVariation = "default";
 
   $: ({ slug, title, subhead, byline, type, publicationDate } = entry);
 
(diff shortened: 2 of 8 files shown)


src/lib/components/Date/Date.scss
```diff
@@ -2,11 +2,17 @@
   display: flex;
   flex-direction: column;
   text-align: center;
-  width: 78px;
-  height: calc(39px * 2 + 2px);
```
LouisFettet, line 6: @Benaiah was there a reason for the additional `2px` here? I adjusted it below so that it renders as a square since `box-sizing` should be set to `border-box`.

src/routes/(infoPages)/about/news/NewsEntrySnippet.svelte
```diff
@@ -20,26 +23,31 @@
...
     <Link href={`/about/news/article/${slug}`}>
       <svelte:element this={headingTagByLevel[headingLevel]} class="news-title">
         {title}
       </svelte:element>
     </Link>
-    {#if subhead}
-      <p>{subhead}</p>
+    {#if variation === "homepage-listed"}
+      <div class="margin-bottom-1"><Tag>{type}</Tag></div>
+    {/if}
```
LouisFettet, line 34: Not crazy about how I have this set up, but I wasn't sure if reordering things using CSS would be any better.  We want the tags at the top of the item when they're listed on a truncated version on the hompage, but at the bottom otherwise.

src/routes/page.scss
```diff
@@ -3,15 +3,6 @@
   $theme-font-type-sans: "public-sans"
 );
 
-.usa-prose > .ldaf-homepage__popular-resources-heading {
-  margin-bottom: 1rem;
-  @include u-font("sans", $theme-h2-font-size);
-}
```
LouisFettet, line 9: This wasn't doing anything since `.ldaf-homepage__populat-resources-heading` is not a direct descendant of `.usa-prose`. I think what I have now roughly matches the mocks, but we still need to do a full-bleed container (for both "Popular resources" and "From the commissioner") to get a closer match.

src/routes/page.scss
```diff
@@ -3,15 +3,6 @@
...
-
-.ldaf-homepage__popular-resources {
-  margin-bottom: 1.5rem;
-}
```
LouisFettet, line 13: Swapped this out for a `margin` class on the element.

LouisFettet (commented):
self-review

PR #479, opened 2023-09-21
[LDAF-431] Create 404 error page

LouisFettet:
Ticket: [LDAF-431](https://ldaf.atlassian.net/browse/LDAF-431)

Our error page is currently very bare-bones, so UX wanted us to at least implement a friendlier page to handle 404s. LDAF expressed an interest in being able to control the error page content, so I created [a new Error CT](https://app.contentful.com/spaces/pc5e1rlgfrov/content_types/error/fields) in Contentful.

## Proposed changes

- Greatly expands our expected error page, including using content fetched from Contentful to populate the heading, subheading, description, image, and meta title. If content for a particular error doesn't exist in Contentful, we output some generic information.
- Loads error page content once when the server starts up (see my self-review comment for more on this), including loading a blurhash for the error image.
- Updates our GraphQL schema and type declarations.

[LDAF-431]: https://ldaf.atlassian.net/browse/LDAF-431?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/routes/+error.svelte
```diff
@@ -1,10 +1,55 @@
...
-  {#if $page.error}
-    <h1>{$page.error.title || $page.status}</h1>
-    <p>{$page.error.message}</p>
-  {/if}
-</div>
+<svelte:head>
+  <meta name="robots" content="noindex nofollow" />
+</svelte:head>
+
+<!-- To reduce further error possibilities, we use as much plain HTML here as possible. -->
```
LouisFettet, line 10: There's a couple things I could change here if we don't think my comment above is an issue:
* optimize the error image and generate a blurhash for it
* use `<Button>` components below (using `isLink`).

Benaiah: I don't think a 404 happening will be particularly associated with front-end JS errors. We would want to avoid as much as possible in a [`handleError`](https://kit.svelte.dev/docs/hooks#shared-hooks-handleerror) function, but a 404 is an "expected error" so I think it's safe to use our normal components.

src/lib/loadErrorPageContent.ts
```diff
@@ -0,0 +1,109 @@
...
+
+import assetProps from "$lib/fragments/assetProps";
+import entryProps from "$lib/fragments/entryProps";
+
+import type { ErrorPageCollectionQuery } from "./$queries.generated";
+import type { ExtractQueryType } from "$lib/util/types";
+
+// TODO: This runs once when the server starts; we should instead run it once per user session.
+let ERROR_CONTENT_LOADED = false;
+const errorPageMap: ErrorPageMap = new Map();
```
LouisFettet, line 16: If y'all have any ideas how we can do this easily, I'm all ears. I tried a few different things but kept running into brick walls. Without individual user session management (which I think is probably out-of-scope here), I'm not sure if there's a way to have this run the way I'd like.

Benaiah (approved):

PR #468, opened 2023-09-20
[LDAF-427] Update colors across the site (links, buttons, etc.)

LouisFettet:
Jira ticket: [LDAF-427](https://ldaf.atlassian.net/browse/LDAF-427)

The Figma mocks for [links](https://www.figma.com/file/oGKbyCnCRRdNzLYbiags93/LDAF-Component-Library-USWDS-3.0.2?type=design&node-id=2178-9666&mode=design&t=cey80GrlG7U7XGB3-4) and [buttons](https://www.figma.com/file/oGKbyCnCRRdNzLYbiags93/LDAF-Component-Library-USWDS-3.0.2?type=design&mode=design&t=cey80GrlG7U7XGB3-0) have been updated after an audit of our colors. This might not cover everything, but it will at least get us closer to the mocks.

## Proposed changes

- Make a handful of color updates based on mocks.
- Utilize the `<Link/>` component wherever it makes sense to do so.
- Override the `usa-link` class colors provided by the USWDS.
- Utilize our color variables in the `<Button/>` component styling, and ensure links that render with button styling utilize the button styling (e.g. no `visited` state). Add a new story for every button variation when it is rendered as link.
- Remove the "secondary" button variation from use. CTA buttons on the homepage, top tier pages, and core content pages should all now show the first CTA with the "primary" variation and all following CTAs with the "outline" variation.
- Provide some default `focus` styles to override the USWDS default outline color. Make a few updates so that focus styles on dark backgrounds use a lighter outline.
- Handle a metric ton of edge cases for setting colors. These are mostly for situations where we need to override the USWDS when it's being overly specific. Some examples include:
  -  the "Return to top" link in the footer 
  - breadcrumb links
  - skip nav link
  - primary and secondary nav links
  - sidenav links
  - special overrides for `focus` outlines on elements in the header at different screen sizes
  - video card links

## Acceptance criteria validation

- [x] Ensure all links are using the `<Link/>` component as appropriate.
- [x] All links (except special cases such as the secondary nav in the header and the links in the bottom half of the footer) should be using the colors outlined on the [typography board in Figma](https://www.figma.com/file/oGKbyCnCRRdNzLYbiags93/LDAF-Component-Library-USWDS-3.0.2?type=design&node-id=2178-9593&mode=design&t=i8ySkXmzN7RwXtom-4).

### Alternate solutions

[...]

PR #489, opened 2023-10-05
[LDAF-427] Update link and button colors across site by using USWDS theming settings

LouisFettet:
Jira ticket: [LDAF-427](https://ldaf.atlassian.net/browse/LDAF-427)

This is a replacement of #468; you don't necessarily need to look at the code on that PR, but the PR description provides some context.

Previously I had run into issues with using the USWDS theming settings because there are some settings that must take tokens (I was under the impression that all setings had to take tokens). I've documented this relatively well in `src/app.scss,` but basically there are two different types of settings: those where you can provide custom colors (we're using hex color variables), and those where you must use tokens. Thankfully in most cases we can use tokens that we're able to customize, but there are some outliers (in this PR, the main one is the visited link color).

## Proposed changes

- Sets all applicable USWDS settings variables to the appropriate LDAF colors. The only setting we currently can't handle properly (at least for links) is `theme-link-visited-color`. Made plenty of notes in-code about why this is.
- Use the `Link` component for all links handled by `ContentfulRichText` component.
- Updates styling on the `Link` component to handle the visited state color (which is ignored in some circumstances) and the inverse variation (when used on a dark background).
- Remove the "secondary" button variation from use. CTA buttons on the homepage, top tier pages, and core content pages should all now show the first CTA with the "primary" variation and all following CTAs with the "outline" variation.
- Makes tons of updates to `Button` component styling. Includes:
  - no-op changes to use available variables
  - updates to get colors to match with mocks
  - removal of a lot of color rules that are no longer needed thanks to using USWDS settings
- Adds a `Button` story for the `secondary` variation.
- Handles some use-cases for a lighter focus outline color for darker backgrounds (like in the footer and desktop header).
- Renames some color variables:
  - `$base-base` is now just `$base`
  - `$disabled` colors were previously `$disabled`, `$disabled-dark`, and `$disabled-darker`, now they're `$disabled-light`, `$disabled`, and `$disabled-dark`
  - `$primary-darkest` is now `$primary-darker`
- For consistency, use the `$white` variable instead of `#fff` or `white`.

- [x] Ensure all links are using the `<Link/>` component as appropriate.
- [x] All links should be using the colors outlined on the [typography board in Figma](https://www.figma.com/file/oGKbyCnCRRdNzLYbiags93/LDAF-Component-Library-USWDS-3.0.2?type=design&node-id=2178-9593&mode=design&t=i8ySkXmzN7RwXtom-4).

## Requested feedback

See self-review.

[LDAF-427]: https://ldaf.atlassian.net/browse/LDAF-427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app.scss
+++ b/src/app.scss
@@ -2,12 +2,103 @@
 @use "uswds" as * with (
   $theme-font-path: $theme-font-path,
   $theme-image-path: $theme-image-path,
+
+  //// Colors - Customizable Theme Colors ////
+  // Some of these color settings are no-ops since our LDAF theme palette has
+  //   some overlap with the default USWDS theme palette.
+  // Our LDAF color variables can be found in src/variables.scss
+  // USWDS default colors can be found in the documentation:
+  //   https://designsystem.digital.gov/design-tokens/color/theme-tokens/
+  // ... with some additional undocumented options in the repo:
+  //   https://github.com/uswds/uswds/blob/v3.4.1/packages/uswds-core/src/styles/settings/_settings-color.scss
+  $theme-color-base-lightest: $base-lightest,
+  $theme-color-base-lighter: $base-lighter,
+  $theme-color-base-light: $base-light,
+  $theme-color-base: $base,
+  $theme-color-base-dark: $base-dark,
+  $theme-color-base-darker: $base-darker,
+  $theme-color-base-darkest: $base-ink,
+  $theme-color-base-ink: $base-ink,
+  $theme-color-primary-lightest: $primary-lightest,
+  $theme-color-primary-lighter: $primary-lighter,
+  $theme-color-primary-light: $primary-light,
+  $theme-color-primary: $primary,
+  $theme-color-primary-vivid: $primary-vivid,
+  $theme-color-primary-dark: $primary-dark,
+  $theme-color-primary-darker: $primary-darkest,
+  $theme-color-secondary-lighter: $accent-lighter,
+  $theme-color-secondary-light: $accent-light,
+  $theme-color-secondary: $accent-warm,
+  $theme-color-secondary-dark: $accent-dark,
+  $theme-color-secondary-darker: $accent-darker,
+  // Skip accent colors since we're not using them. Available options are:
+  // $theme-color-accent-cool-lighter
+  // $theme-color-accent-cool-light
+  // $theme-color-accent-cool
+  // $theme-color-accent-cool-dark
+  // $theme-color-accent-cool-darker
+  // $theme-color-accent-warm-lighter
+  // $theme-color-accent-warm-light
+  // $theme-color-accent-warm
+  // $theme-color-accent-warm-dark
+  // $theme-color-accent-warm-darker
+  $theme-color-info-lighter: $info-lighter,
+  $theme-color-info-light: $info-light,
+  $theme-color-info: $info,
+  $theme-color-info-dark: $info-dark,
+  $theme-color-info-darker: $info-darker,
+  $theme-color-error-lighter: $error-lighter,
+  $theme-color-error-light: $error-light,
+  $theme-color-error: $error,
+  $theme-color-error-dark: $error-dark,
+  $theme-color-error-darker: $error-darker,
+  $theme-color-warning-lighter: $warning-lighter,
+  $theme-color-warning-light: $warning-light,
+  $theme-color-warning: $warning,
+  $theme-color-warning-dark: $warning-dark,
+  $theme-color-warning-darker: $warning-darker,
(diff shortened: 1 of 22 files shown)


src/lib/components/Alert/Alert.scss
```diff
@@ -3,8 +3,6 @@
   border: none;
 }
 .ldaf-alert--site-wide.usa-alert--error {
-  background-color: $error-dark;
```
LouisFettet, line 6: Pretty sure this had no effect.

src/lib/components/Alert/Alert.scss
```diff
@@ -15,16 +13,8 @@
...
     }
   }
 }
-
-.ldaf-alert--site-wide.usa-alert--info .usa-alert__body .usa-link {
-  color: $primary-light;
-  &:visited {
-    color: $primary-light;
-  }
-}
```
LouisFettet, line 30: Link color is now set via the settings. Talked with Courtney and he said it should be fine to show the visited state on alerts (with the caveat that we want to keep the link white on the site-wide emergency alert).

src/lib/components/Button/Button.scss
```diff
@@ -1,24 +1,9 @@
...
-  "disabled-dark": #adadad,
-  "base-lightest": #f2f1f0,
-  "base-light": #a9aeb1,
-  "base-dark": #565c65,
-  "base-darker": #3d4551,
-  "gray-05": #f0f0f0,
-  "primary-lightest": #d1e9ff,
   "text-only-hover": rgba(6, 60, 122, 16%),
   "text-only-active": rgba(6, 60, 122, 30%),
-  "outline-inverse": #dcdee0,
```
LouisFettet, line 21: `outline-inverse` was removed since it wasn't being used after we corrected some things to use `base-light` instead.

src/lib/components/Button/Button.scss
```diff
@@ -57,104 +44,84 @@ $filters: (
...
+      filter: map.get($filters, "white");
+    }
   }
 }
 
 .usa-button.usa-button--base {
-  background: map.get($colors, "base-dark");
-  &:hover {
-    background: map.get($colors, "base-dark");
-  }
```
LouisFettet, line 101: Hover state on the base button should show no change. Adding a rule here would be redundant.

src/lib/components/VideoCard/VideoCard.scss
```diff
@@ -1,7 +1,3 @@
-@use "uswds-core" as * with (
-  $theme-show-notifications: false
-);
```
LouisFettet, line 3: I think this was leftover from previous work that utilized USWDS mixins. Should no longer be required.

LouisFettet (commented):
self-review!

Benaiah (approved):

src/app.scss
```diff
@@ -2,12 +2,103 @@
...
   $theme-font-role-heading: "sans",
   $theme-font-role-body: "sans",
   $theme-font-role-alt: "sans",
-  $theme-show-notifications: false
+
+  //// Logs Settings ////
+  // Turn off notifications to avoid noise in the logs.
+  $theme-show-notifications: false,
+  // Turn off compilation warnings due to excessive contrast a11y warnings.
+  $theme-show-compile-warnings: false
```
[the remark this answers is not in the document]
hinzed1127: > This is really unfortunate, but our logs just get absolutely hammered...
I agree, but it's probably not worth all the noise, and we can always uncomment if for some reason we want to examine other warnings

src/app.scss
```diff
@@ -2,12 +2,103 @@
...
+  // $theme-color-accent-cool-lighter
+  // $theme-color-accent-cool-light
+  // $theme-color-accent-cool
+  // $theme-color-accent-cool-dark
+  // $theme-color-accent-cool-darker
+  // $theme-color-accent-warm-lighter
+  // $theme-color-accent-warm-light
+  // $theme-color-accent-warm
+  // $theme-color-accent-warm-dark
+  // $theme-color-accent-warm-darker
```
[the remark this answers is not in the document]
hinzed1127: IMO I like explicitly noting what's not getting used. It's not that many, and useful to note in case there's a desire to change it in the future.

src/app.scss
hinzed1127, line 1: Thank you for all of the additional commentary here. Very helpful and clear in what's being done!

hinzed1127 (approved):
Thanks for all of the work/re-work in getting this updated!

PR #502, opened 2023-10-14
Move Contentful client initialization to hook

Benaiah:
## Proposed changes

- Moves contentful client code into server-only folder
- Moves the contentful client initialization to a single place in a server hook
- Migrates all places where the contentful client is initialized to use the client from the hook

- [x] Fulfilled acceptance criteria
- [x] Existing tests continue to pass
- [x] Manually tested


src/lib/services/server/contentful/graphqlClient.ts
```diff
@@ -39,14 +40,15 @@ const getClient = ({
...
+  fetch = global.fetch,
   apiPrefix = defaultAPIPrefix,
 }: ClientOptions): Client => {
   const key = getKeyFromOptions({ spaceID, environment, token });
   const existingClient = clients.get(key);
   if (existingClient) return existingClient;
 
   const client = {
-    options: { spaceID, environment, token, apiPrefix },
+    options: { spaceID, environment, token, apiPrefix, fetch },
```
hinzed1127, line 51: what's the need to have fetch be a parameter?

hinzed1127 (approved):
small clarifying question but otherwise looks good

Benaiah:
> small clarifying question but otherwise looks good

There wasn't one, good catch. I've removed that from the PR.

PR #511, opened 2023-10-27
Allow logging in and out of site with personal access token

Benaiah:
## Proposed changes

- Allows logging into and out of the site using a personal access token.

- [x] You can log into the site with a personal access token
- [x] You can log out of the site
- [x] All existing functionality continues to work
- [x] Tests added and existing tests pass


src/routes/logout/+page.server.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const actions = {
+  default: async ({ cookies, locals: { getKVClient } }) => {
+    const userToken = cookies.get("ldafUserToken");
+    if (!userToken) return { success: true };
+    cookies.delete("ldafUserToken", { sameSite: "none", path: "/" });
+    try {
+      const kvClient = await getKVClient();
+      await kvClient.deleteUserInfoByToken(userToken);
+    } catch (err) {
```
hinzed1127, line 11: So if I just navigate to `/logout` when I'm already logged in, it continues to say I'm logged in until  I click the "logout" button. However, it looks like the related redis values and user token cookies are getting deleted here. Are they getting instantiated elsewhere on page load? I'm trying to wrap my head around the difference between fully logging out and loading this page url by itself (not a likely path to occur, I know, but humor me).

src/routes/logout/+page.svelte
```diff
@@ -0,0 +1,26 @@
...
+  $: if (form?.success) currentUser.set(undefined);
+</script>
+
+<div class="grid-container">
+  <h1>Logout</h1>
+  {#if $currentUser}
+    {#if form?.success}
+      <p>Successfully logged out!</p>
+    {:else}
+      <p>You are logged in as {$currentUser.name}.</p>
```
hinzed1127, line 18: this is related to my question in the complementary `+page.server.ts` file for this route

hinzed1127 (commented):
This looks pretty good overall. I have a few stray observations that may or may not require a small amount of cleanup, but should be good to go in after that. Appreciate the thorough testing here, too!

hinzed1127 (approved):

PR #520, opened 2023-11-10
Allow viewing preview content while logged in

Benaiah:
## Proposed changes

Allows viewing preview changes while logged in by adding `?preview` to the URL.

- [x] Fulfilled acceptance criteria
- [x] Manually tested


src/routes/+layout.svelte
```diff
@@ -69,6 +72,40 @@
...
+    if (
+      typeof previousPreviewValue === "string" &&
+      type === "link" &&
+      to &&
+      !to.url.searchParams.has("preview") &&
+      !lastPageError &&
+      !previewAuthenticationError &&
+      !(to.url.pathname === "/login" || to.url.pathname === "/logout")
+    ) {
+      to.url.searchParams.set("preview", previousPreviewValue);
```
hinzed1127, line 91: small nit:  because of how `searchParams.set()` functions, all subsequent page loads include an "=" at the end, e.g. `ldaf.la.gov/some-sub-route?preview=`. It doesn't look like it would take a lot to update logic to use `preview=true` and then check for that, i.e., also prevent the confusing state of showing preview content even if the url has `?preview=false` at the end

src/routes/+layout.svelte
```diff
@@ -123,7 +160,18 @@
...
     bind:navMenuExpanded
   />
-  <slot />
+  {#if previewAuthenticationError}
+    <div class="usa-section usa-prose grid-container">
+      <h1>{previewAuthenticationError.status}</h1>
+      <p>{previewAuthenticationError.message}</p>
+      <p>
+        (Do you need to <LoginLink {...loginLinkProps}>log in</LoginLink> or visit the
+        <a href={$page.url.toString()}>public version of the page</a>?)
```
hinzed1127, line 169: `$page.url` still includes `?preview` at the end of it. That needs to get stripped out to make sure it's a functional link.

hinzed1127 (commented):
Looks good!

I think the logic in `src/lib/components/LoginLink/LoginLink.svelte` needs to updated to handle navigation after a successful login. Right now the `state` parameter` doesn't seem to be getting used to successfully navigate to the proper preview page post-authentication.

Other than that, just a few small things. The most important scenarios seem to all be functioning correctly.

Benaiah:
@hinzed1127 thanks for the review! addressed your comments and updated the PR

src/routes/login/+page.svelte
```diff
@@ -21,12 +25,25 @@
...
+  }
 </script>
 
 <div class="grid-container">
   {#if form?.success}
     {#if form.currentUser}
       <p>Successfully logged in!</p>
+      {#if redirectURL}
+        <p>Please wait to be redirected to the previous page.</p>
+      {/if}
```
hinzed1127, line 46: Nit: In an ideal situation, this would hopefully be visible for a very short amount of time.  A UX improvement could be to have this appear in an accessible modal, along with providing a `redirectURL` link in case the redirect doesn't happen for some reason. I'd leave that for follow-up work for now.

hinzed1127 (approved):
Everything is working now for me! LGTM

src/routes/login/+page.svelte
Benaiah: This is already visible for a very short amount of time - the `await timeout(500)` above does that. I've added a manual redirect link in case that doesn't work.

PR #525, opened 2023-11-15
Allow for logging in with Contentful OAuth

Benaiah:
## Proposed changes

- Allows logging into the site with an OAuth2 flow that uses your existing Contentful login

- [x] Manually tested
- [x] Existing tests pass

## Other details

Post-merge checklist:

- [ ] Change environment variables in Vercel to point at the production URL of the LDAF site
- [ ] Change OAuth2 app settings in Contentful to point at the production URL of the LDAF site


src/lib/components/ContentfulLoginLink/ContentfulLoginLink.svelte
```diff
@@ -0,0 +1,38 @@
+<script lang="ts">
+  import type { ComponentProps } from "svelte";
+  import {
+    PUBLIC_CONTENTFUL_OAUTH_ENDPOINT,
+    PUBLIC_CONTENTFUL_OAUTH_CLIENT_ID,
+    PUBLIC_CONTENTFUL_OAUTH_CLIENT_REDIRECT_URI,
+  } from "$env/static/public";
+  import { page } from "$app/stores";
+  import Link from "$lib/components/Link";
+  import classNames from "$lib/util/classNames";
```
hinzed1127, line 10: Linting cleanup

src/routes/callback/oauth2/+page.svelte
```diff
@@ -0,0 +1,99 @@
...
+    }
+  };
+
+  if (browser) handleHash();
+</script>
+
+<div class="grid-container">
+  {#if loginErrorMessage}
+    <p>Failed to log in!</p>
+    <p>{loginErrorMessage}</p>
```
hinzed1127, line 85: I don't think `loginErrorMessage` ever gets initialized to a value where this will display

hinzed1127 (approved):
A couple small things and then good to merge.

I know w've got login/logout tests, but a good follow-up would be to add some tests specific to the OAuth login flow. That's probably more suited to puppeteer, I'd think?

PR #556, opened 2024-05-30
Replace ts-node with tsx to resolve errors running pre scripts.

LouisFettet:
I'm still not entirely sure why these scripts started failing on me, but it seems like `ts-node` was causing issues. After replacing all the uses / configurations of `ts-node` with `tsx` (based on [this SO answer](https://stackoverflow.com/a/76343394)), everything is working for me again locally. 

## Proposed changes

- Replace `ts-node` dependency with `tsx` and update the `scripts` in `package.json` to use `tsx`.
- Remove the various minor configurations related to `ts-node` (`tsx` touts "No configuration required!" on [its site](https://tsx.is/).

### Alternate solutions

It's possible that updating `ts-node` could have fixed the issue. Didn't try this. I think the fact that it just stopped working out of nowhere is a bit concerning though, and `tsx` seems to be gaining adoption so it seems like the safer bet.

## Requested feedback

@hinzed1127 let me know if this works for you. I'm also curious if you were you running into any issues with running the pre scripts as well. I'm a bit confused why it would only impact me.


hinzed1127 (approved):
I was encountering the same TypeScript errors and this fixed it. Switching over to `tsc` makes sense to me 👍

PR #557, opened 2024-05-30
Implement internal and external redirects using PageMetadata from Contentful.

LouisFettet:
We had originally built the `Page details` content type with support for internal redirects (a reference to another `Page details` entry) and external redirects (a URL string), but we never implemented the feature on the site. 

## Proposed changes

- Split our query for the `pageMetadataMap` based on the need for redirects. Change our default parameters for `loadPageMetadataMap` so that we assume we don't need breadcrumbs or redirects. Include the data for redirects in the query, if available.
- In our base layout server file (`src/routes/+layout.server.ts`), we attempt to find a matching `Page details` entry for the requested path, check if that entry has any internal or external redirects, and `throw` an error using the built-in SvelteKit `redirect` method.
- In Contentful, update the internal redirect field on `Page details` so that (in addition to another `Page details` entry) it will also accept a `News` or `Event` entry. Support all three types of internal redirects in the base layout server file.

To test this out locally, switch your `DEFAULT_CONTENTFUL_ENVIRONMENT` in your `.env.local` config file from `master` to `redirect-dev` (I set up this env in Contentful to avoid causing any problems on production).

- [x] Navigating to `/indian-creek-recreation-area/book-your-stay-today` **externally** redirects to [the ResNexus site](https://resnexus.com/resnexus/reservations/book/CF0B092F-8EF5-409F-A4EE-231FDD506F13)
  - `Page details` entry: [Book your stay at Indian Creek](https://app.contentful.com/spaces/pc5e1rlgfrov/environments/redirect-dev/entries/4KOsBcTfpy3XwEtdocvZgO) (This entry was already set up with an external redirect, so no changes were made here.)
- [x] Navigating to `/indian-creek-recreation-area/getting-to-indian-creek` **internally** redirects to `/indian-creek-recreation-area/activities-amenities`.
  - `Page details` entry: [Getting to Indian Creek](https://app.contentful.com/spaces/pc5e1rlgfrov/environments/redirect-dev/entries/6AG8EtY2t9yKt7JgEXo6cv) (This is not a real use-case and was just added for testing.)
- [x] Navigating to `/online-payments` **internally** redirects to `/business/online-payments`.
  - `Page details` entry: [Horticulture Online Payments](https://app.contentful.com/spaces/pc5e1rlgfrov/environments/redirect-dev/entries/5ioPmhRMcMJBMFU9LEehGW) (This entry was already set up with an internal redirect, so I just set the parent page to the homepage for testing.)
- [x] Navigating to `/indian-creek-recreation-area/campground-maps` **internally** redirects to `/about/events/event/2024-05-29-louisiana-agricultural-commodities-commission-meeting`.
  - `Page details` entry: [Indian Creek Maps](https://app.contentful.com/spaces/pc5e1rlgfrov/environments/redirect-dev/entries/1IqbtOPhCbyk6AUiPc3bLE) (For testing only.)
- [x] Navigating to `/indian-creek-recreation-area/pavilion-reservations` **internally** redirects to `/about/news/article/ag-commissioner-mike-strain-in-cuba-building-bridges-for-la-export`.
  - `Page details` entry: [Pavilion reservations](https://app.contentful.com/spaces/pc5e1rlgfrov/environments/redirect-dev/entries/1yje6Io98nE94zhWsBRTSt) (For testing only.)
- [x] Pages that redirect do not appear in the mega menu or the side nav.
 

### Alternate solutions

While it might have been nice to handle this in a `hook` instead, I wanted to keep things relatively simple and add the logic after we load the `pageMetadataMap` (I'm honestly not sure how this would look in a `hook` unless we built the map in the hook, which doesn't seem right.).

I originally added some of this logic in the individual `+page.server.ts` files for the `[topTierPage]` and `[...serviceGroupPage]` routes, but this was repetitive, and I think handling it in the base layout is cleaner. We are doing a bit of double work now that we're looking up the `Page details` entry via the map twice (once in the base layout to determine a redirect and once again when loading Top Tier and Core Content pages), so this could probably still be improved.

### Possible drawbacks

This setup allows for redirecting from any path (both those that already exist and ones that don't exist yet), but the editor experience for redirecting from a non-existent path (especially a nested one) is a bit weird. For instance, if you wanted to redirect from a new nested path `/abc/xyz` to `/`, you would need to set up entries for both `abc` and `xyz` and add the relevant redirect on `xyz`. If `abc` is just there for the URL, you will need to add a redirect on `abc` as well so that it doesn't show up as a weird error page.

Additionally, there doesn't seem to be an easy way to prevent an infinite redirect loop; if two pages redirect to each other, the site will likely break pretty badly. We could add some basic protection against this in the base layout server file, but it would decrease efficiency and still be relatively difficult to handle complex loop situations (e.g. `/a` redirects to `/b`, `/b` redirects to `/c`, and `/c` redirects to `/a`). Rather than trying to protect against this edge case, it's probably wiser to alert editors to this edge case in the user guide.

## Requested feedback

I'll leave a self-review, but I'm rusty, so any and all feedback here is appreciated!

## To-Do

Some things we should probably think through / update:
* [x] Figure out how to handle paths with redirects in the sitemap and search.
  * Opted to filter out paths with redirects from the sitemap and Algolia search index. I think this is the right call, as these search results likely wouldn't be useful. We can ask LDAF folks if they feel differently before we deploy.
* [ ] See about cutting down on repetition in querying the `pageMetadataMap` in both the base layout and individual page server files (it's only loaded from Contentful once, but the extra lookups are wasteful).
  * Opted not to handle this here, but left a `TODO` comment in the base layout server file.
* [x] Figure out how to handle internal redirects to pages without `Page details` entries such as news and events.

Before merging / deploying, we'll want to:
* [x] Add News and Events as supported entry types for internal redirects on Page Details.
* [ ] Delete the `redirect-dev` environment.
* [x] Write up a quick how-to guide for adding redirects. Should include how to set up internal and external redirects for pages that do and don't exist (with additional instructions for nested paths), and in all cases should account for the weirdness around News and Events not having Page Details associated with them.
* [x] Check with LDAF folks that things are set up the way they want (e.g. see if they want `/online-payments` to internally redirect to `/business/online-payments` and `/indian-creek-recreation-area/book-your-stay-today` to externally redirect to ResNexus).
* [ ] Publish the Page Details entries for the intermediary slugs under News and Events:
  * [News Parent Redirect](https://app.contentful.com/spaces/pc5e1rlgfrov/environments/master/entries/51RdGXuKdA2iQ8xMlyJI76) (for the “article” slug)
  * [Event Parent Redirect](https://app.contentful.com/spaces/pc5e1rlgfrov/environments/master/entries/kX46qBuNEqAkXCm1htmbR) (for the “event” slug)


src/lib/loadPageMetadataMap.ts
```diff
@@ -39,6 +39,14 @@ const query = gql`
             }
           }
         }
+        internalRedirect {
+          ... on PageMetadata {
+            sys {
+              id
+            }
```
LouisFettet, line 46: Most likely will need to add some additional queries here for News and Events.

src/routes/+layout.server.ts
```diff
@@ -11,39 +12,59 @@ export const load = async ({
...
+  const metadataID = pathsToIDs.get(pathname);
+  if (metadataID) {
+    const pageMetadata = pageMetadataMap.get(metadataID);
+    if (pageMetadata) {
+      if (pageMetadata.internalRedirect) {
+        const internalRedirectPageMetadata = pageMetadataMap.get(
+          pageMetadata.internalRedirect.sys.id,
+        );
+        if (internalRedirectPageMetadata?.url) {
+          throw redirect(301, internalRedirectPageMetadata.url);
```
LouisFettet, line 39: `301` felt like the right option here for the ones available, but let me know if you think there's a better option.

LouisFettet (commented):
Quick initial self-review.

src/routes/api/v1/metadata-map/+server.ts
```diff
@@ -5,6 +5,7 @@ import { loadPageMetadataMap } from "$lib/loadPageMetadataMap";
 export const GET = (async ({ locals: { contentfulClient } }) => {
   const { pageMetadataMap } = await loadPageMetadataMap({
     includeBreadcrumbs: false,
+    includeRedirects: false,
```
LouisFettet, line 8: Unsure if this should go here; I think this API endpoint was added after I left? I'm unsure how / where it's being used.  Does this help power search, preview, or something else?

hinzed1127: If I recall correctly, this endpoint was just added as a utility endpoint of sorts as a way to quickly view the metadata map via a request. I think I used it to assist building out search. It's not an active dependency for any other feature. I think it's a nice thing to keep around as a quick way to check those details when needed.

hinzed1127 (commented):
I think this is a totally viable solution. The only alternative approach I can immediately think of is having a separate metadata query that's much slimmer and _only_ grabs redirects. Or in a similar vein, just making the `includeRedirects` and `includeBreadcrumbs` named arguments default to false in the function definition, as to avoid accruing too much bloat in the function arguments, but that concern might still be a little premature.

src/routes/api/v1/metadata-map/+server.ts
LouisFettet: Ah okay, that makes sense. Thanks for the explanation! I don't think it's harming anything, so I'm cool with keeping it around.

LouisFettet:
Got redirects to News and Events pages mostly working. Now noticing that my workaround to redirect **from** a News page has added a bunch of junk to the sidebar, which we don't want. Will look into this a bit more tomorrow.

src/lib/loadPageMetadataMap.ts
```diff
@@ -22,27 +22,65 @@ export type PageMetadataMapItemWithObjectID = PageMetadataMapItem & {
...
-        title
-        slug
-        isRoot
-        parent {
-          ... on PageMetadata {
-            sys {
-              id
-            }
+      }
+      internalRedirect {
```
LouisFettet, line 41: RE: @hinzed1127 separating the query so we only grab redirects, I think that would have been less efficient since we would be making double the requests to Contentful (when we could just grab it all in one go while we're working on loading the PageMetadataMap).

That said, I wanted to create a fragment so we could have two separate queries (one for all the entries and one for only entries that don't include redirects), but in both cases here we're still asking for data on `internalRedirect` and `externalRedirect` even though in one of those cases we know that these will all come back `null`. Unsure how to clean this up so that we don't query for values we know are `null` without removing the fragment and duplicating the majority of the query.

src/lib/loadPageMetadataMap.ts
```diff
@@ -98,9 +136,14 @@ const constructBreadcrumbs = (
 };
 
 export const loadPageMetadataMap = async ({
-  includeBreadcrumbs = true,
+  includeBreadcrumbs = false,
+  includeRedirects = false,
```
LouisFettet, line 140: Changed these to default to `false` per your suggestion @hinzed1127. Even though the main use-case includes these, we have more use-cases where we don't.

src/lib/loadPageMetadataMap.ts
hinzed1127: > Unsure how to clean this up so that we don't query for values we know are null...

Not sure if I follow this bit. What scenario would do that?

hinzed1127 (approved):
👍 Looks good! And honestly the end result is pretty straightforward, just the small wrinkle to account for news & events and press releases separately

src/lib/loadPageMetadataMap.ts
LouisFettet: The fragment here is querying for `internalRedirect` and `externalRedirect`, but we use the fragment for both types of queries:
1. `queryAll` - we do this when `includeRedirects` is `true` and therefore we want the data on `internalRedirect` and `externalRedirect`
2. `queryWithoutRedirects` - we do this when `includeRedirects` is `false` and we apply a filter:
    ```
    where: { AND: [{ internalRedirect_exists: false }, { externalRedirect_exists: false }] }
    ```
    But because we're reusing the same fragment we'll still get back `internalRedirect` and `externalRedirect` in the response (all of which will be `null` based on the filter).

Ideally we wouldn't query for `internalRedirect` and `externalRedirect` in the second case, but I don't think there's a way to do that without duplicating the GraphQL query?

PR #559, opened 2024-06-18
Add video embed to core content pages

hinzed1127:
## Proposed changes

- Adds the same video embed that's currently on top tier pages to core content pages. This makes a video embed possible on almost every page, sans News/Events and I think office pages

## Screenshots

Video embed on a core content page
Screenshot 2024-06-18 at 1 46 50 PM

To test this out locally, switch your `DEFAULT_CONTENTFUL_ENVIRONMENT` in your `.env.local` config file from `master` to `video-on-all` (I set up this env in Contentful to avoid causing any problems on production).

- [x] Navigate to `/land/fire` and a video feature LDAF Wildland firefighters should be visible
- [x] Other core content pages without videos still render normally

## Other details

This is a pretty straightforward rip of the video implementation for the top tier pages

### Possible drawbacks

I thought about abstracting the logic to be shared between top tier and core content pages, but that felt a bit premature and makes this addition more removable in the future 

## Requested feedback

I'm getting weird `code-gen` schema errors again. In this case, it's this:
```
✖ Unable to validate GraphQL document!
  File /Users/danhinze/work/ldaf-site/src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts caused error:
  Unable to find field "video" on type "ServiceGroup"!
```

### DIFF
--- a/src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.svelte
+++ b/src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.svelte
@@ -30,6 +30,8 @@
     childServiceGroups,
   } = data);
 
+  console.log(data);
+
   afterUpdate(() => {
     // Call To Actions within service entry accordions are rich text,
     // leaving no way to designate additional formatting/styling for them.
--- a/src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.svelte
+++ b/src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.svelte
@@ -9,9 +9,11 @@
   import ContactCard from "$lib/components/ContactCard";
   import ContentfulRichText from "$lib/components/ContentfulRichText";
   import Icon from "$lib/components/Icon";
+  import VideoCard from "$lib/components/VideoCard";
   import Image from "$lib/components/Image";
   import Link from "$lib/components/Link";
   import { url as arrowIcon } from "$icons/arrow_forward";
+  import serviceGroupPageTestContent from "./__tests__/serviceGroupPageTestContent";
 
   export let data;
   $: ({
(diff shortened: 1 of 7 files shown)


src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.svelte
```diff
@@ -9,9 +9,11 @@
   import ContactCard from "$lib/components/ContactCard";
   import ContentfulRichText from "$lib/components/ContentfulRichText";
   import Icon from "$lib/components/Icon";
+  import VideoCard from "$lib/components/VideoCard";
   import Image from "$lib/components/Image";
   import Link from "$lib/components/Link";
   import { url as arrowIcon } from "$icons/arrow_forward";
+  import serviceGroupPageTestContent from "./__tests__/serviceGroupPageTestContent";
```
Comment (hinzed1127) [inline src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.svelte:16]: TODO: clean up comments and remove this accidental auto-import

LouisFettet (commented):
Looks good overall! Although I'm having trouble remembering whether they wanted to be able to embed videos within RichText, or just have a featured video on pages outside the homepage and top tier pages? Regardless this is a good first step which likely accomplishes most of their needs (I don't think showing multiple video embeds on a page will be a common use-case).

Got two notes that don't fit well into comments on the code:
1. I was able to get the schema error resolved by setting the `CONTENTFUL_DEFAULT_ENVIRONMENT` to `video-on-all` in *_both_* the `.env.local` and `.env` files. It seems that the schema generation scripts always use whatever's in `.env`? Honestly I found this through pure luck because on my redirect work I had tried setting `.env` first before remembering I should use `.env.local`.
2. I just noticed that we have a styling error on the subscribe button when the link is unvisited; the `usa-link` styles override the link's `color` on `:hover`. As a quick fix I think you can just add another selector for `.usa-link.youtube-subscribe-link:hover` here: https://github.com/la-ldaf/ldaf-site/blob/51ace3822f3b2a640d73592179102b71b400a9ff/src/lib/components/YoutubeSubscribeLink/YoutubeSubscribeLink.scss#L4

hinzed1127:
> I was able to get the schema error resolved by setting the CONTENTFUL_DEFAULT_ENVIRONMENT to video-on-all in both the .env.local and .env files. 

That did it!

> I'm having trouble remembering whether they wanted to be able to embed videos within RichText, or just have a featured video on pages outside the homepage and top tier pages?

From a UX/established patterns perspective, I'd rather gradually enable adding videos rather than open it up to anywhere and everywhere. This should hopefully also cover the majority of other cases a video might be appropriate. We can always tack onto this as necessary.

LouisFettet (approved):
LGTM! I forget if I mentioned this previously, but I also added the same video embed to a Core Content page without featured services, `/land/fire/safety`. The placement below the main RichText content makes sense to me.

Here's what I think we need to do to get this out the door:
* Before merging...
  * Merge changes in from `main`
  * Make the appropriate changes to the content models in the `master` environment in Contentful (i.e. add the "Video" field to the "Core Content" model).
  * Switch your `.env` to use the Contentful `master` environment and run `gen-schema` and update the Contentful GraphQL schema files.
* After merging...
  * Delete the `video-on-all` environment in Contentful.
 
 Is there anything I'm missing above?

hinzed1127:
> I forget if I mentioned this previously, but I also added the same video embed to a Core Content page without featured services, /land/fire/safety.

Yup, I validated that it rendered there correctly, too. Looks good to me.

> Is there anything I'm missing above?

Nope, I think that covers it! I can run through all the Contentful stuff today and get us in a position to hopefully merge this, too.

PR #560, opened 2024-07-09
Block bad actors by immediately responding with an HTTP error code.

LouisFettet:
Before we perform any load logic, we check the user agent of the request and give them an error if we know they're up to no good. More details are available in the in-code comments.

### Testing

Open up your HTTP-request-tool of choice and send a request to any URL on the site (either to your local or to the Vercel preview deployment) with the `User-Agent` header set to:
```
FeedFetcher-Google; (+http://www.google.com/feedfetcher.html)
```
All requests (besides to static assets such as `/robots.txt` and `/favicon.ico`) should result in a `419 Never Gonna Give You Up` error. This is not a real error code, but these requests don't deserve a real error code (in all seriousness, I can change this to a `403 Forbidden` if we want to be proper).

### DIFF
--- a/src/hooks.server.ts
+++ b/src/hooks.server.ts
@@ -1,10 +1,12 @@
 import { sequence } from "@sveltejs/kit/hooks";
 
-import handleSetupKVClient from "./hooks/server/handleSetupKVClient";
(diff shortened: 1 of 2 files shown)


hinzed1127 (approved):
🫡

PR #562, opened 2024-08-14
Add news and events to the sitemap.

LouisFettet:
During initial development, we didn't include news and events in search indexing (both via Algolia for in-site search and via the sitemap for external search providers). We're correcting this now for the external search case by adding news and events to the sitemap.

## Proposed changes

- Add new helper load functions for fetching news and events with basic data (only what's needed to provide to the sitemap).
  - Events include the event date/time and slug (both of which are needed to construct the URL).
  - News articles / press releases include the title, publication date, and slug. While the slug is the only thing necessary for a regular sitemap entry, we're including an additional `<news:news>` tag so that these pages will register with news feeds (so that it will show up in something like Google News).
- Add new helper function to construct the slug for event pages (this was previously a repeated line in a few different places).
- Update the sitemap server file to use the new helper load functions and provide news and events in the XML file.

### Alternate solutions

The main alternate solution would be to refactor news and events so that all these queries are in one place (among other clean-up work), but with the limited time available, I think this gets the job done without overcomplicating things. My hope is that the `loadEvents` and `loadNews` helper load functions will be reusable outside of the sitemap.

## Requested feedback

See self-review!


src/lib/loadEvents.ts
```diff
@@ -0,0 +1,33 @@
...
+import { print as printQuery } from "graphql";
+
+import type { ContentfulClient } from "$lib/services/server/contentful";
+import type { EventCollectionQuery } from "./$queries.generated";
+
+type EventItem = NonNullable<EventCollectionQuery["eventEntryCollection"]>["items"][number];
+
+const query = gql`
+  query EventCollection($preview: Boolean = false) {
+    # TODO: A limit of 500 works until it doesn't. We should set up a way to paginate.
```
LouisFettet, line 11: FWIW there are currently `54` events in Contentful.

src/lib/loadNews.ts
```diff
@@ -0,0 +1,34 @@
...
+import { print as printQuery } from "graphql";
+
+import type { ContentfulClient } from "$lib/services/server/contentful";
+import type { NewsCollectionQuery } from "./$queries.generated";
+
+type NewsItem = NonNullable<NewsCollectionQuery["newsCollection"]>["items"][number];
+
+const query = gql`
+  query NewsCollection($preview: Boolean = false) {
+    # TODO: A limit of 500 works until it doesn't. We should set up a way to paginate.
```
LouisFettet, line 11: ... and there are currently `170` news articles / press releases in Contentful.

src/routes/(infoPages)/about/events/shared.server.ts
```diff
@@ -11,7 +11,6 @@ import { getCurrentDateInTZ, getStartOfDayForDateInTZ } from "$lib/util/dates";
 const eventsBasePath = "/about/events";
 const limit = 20;
 
-// TODO: filter to only future events. this will require thinking about timezones
```
LouisFettet, line 14: This was handled in #378, so we can remove this TODO comment.

src/routes/(infoPages)/about/events/constructEventSlug.ts
LouisFettet, line 1: Considered putting this under `src/lib/util` but I think most of those utilities are pretty generic whereas this is quite specific to our routing setup.

larsmaxwell: It might be helpful to have this in utils because it could be used later with algolia search expansion.

larsmaxwell (approved):
I had one suggestion, but LGTM!

src/routes/(infoPages)/about/events/constructEventSlug.ts
LouisFettet: Yeah, ideally utils are kept generic so that they're completely uncoupled from content models and such, but I guess we already have `src/lib/util/getErrorMessageFromResponse.ts` interacting with Contentful errors. Since there isn't really a "best" place for this to fit, I'll move it into utils.

src/lib/components/Event/Event.svelte
```diff
@@ -37,7 +38,7 @@
       <Link
         class="display-block"
         href={date && event?.slug
-          ? `/about/events/event/${date.toISOString().split("T")[0]}-${event.slug}`
+          ? `/about/events/event/${constructEventSlug(date, event.slug)}`
           : undefined}
```
larsmaxwell, line 42: This wasn't created by you, but just something that may be a matter of opinion for setting things as `undefined`. From what I've heard, setting `undefined` vs using `null` is not recommended because its supposed to represent a value that is unassigned. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined

src/lib/loadEvents.ts
hinzed1127: combined with the fact that this is ordered by datetime, I think this should hold up for a while in either case

hinzed1127 (approved):
looks good!

PR #563, opened 2024-09-10
Add API route to process fire data CSV

hinzed1127:
## Proposed changes

- Adds an API route at `/api/v1/fire-weather`
- Presently just parses out an CSV file and returns it as JSON; will get updated to get added to the Vercel KV store

## Other details

I tested this locally with the [following script](https://gist.github.com/hinzed1127/4f51bef0853f6767fb7d7cf8912dffb1). It's intentionally very basic right for now

### Possible drawbacks

- Right now is requiring we integrate with LDAF's IIS FTP server. TBD on how integrating a node script within that goes
- TBD on how easy the KV integration is

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,7 @@
         "@poppanator/sveltekit-svg": "^4.1.0",
         "@uswds/uswds": "^3.4.1",
         "algoliasearch": "^4.20.0",
+        "csv-parse": "^5.5.6",
         "date-fns": "^2.30.0",
         "date-fns-tz": "^2.0.0",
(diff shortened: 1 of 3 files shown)


LouisFettet (approved):
Did a quick test locally, LGTM! Appreciate the incremental approach here, and it should make it easier to continue working with Shawn on this.

Only feedback right now is that it might be good to stub out or leave a comment for better error handling. My guess is that the data won't always be well-formed, and it would be good to provide feedback on exactly what went wrong during parsing. We can handle this in the next one though, so don't let this block you from merging.

PR #571, opened 2024-09-27
Update Svelte v1 - 2

larsmaxwell:
## Closes #12345 

Jira ticket: [TICKET_NUMBER](URL)

## Proposed changes

- 
-

- [ ] 
- [ ] 

## Requested feedback


src/app.scss
```diff
@@ -132,4 +132,7 @@ body {
   & > ol li {
     max-width: none;
   }
-}
+  h1+*, h2+*, h3+*, h4+*, h5+*, h6+* {
```
larsmaxwell, line 135: For some reason, this style is stripped when I ran the upgrades.

src/lib/__tests__/matchers/dom.ts
```diff
@@ -1,3 +1,4 @@
+import type { MatcherFunction } from "@testing-library/svelte";
```
larsmaxwell, line 1: Remove unused import

src/lib/util/resolveWithStatus.server.ts
```diff
@@ -10,7 +10,7 @@ import type { MaybePromise, RequestEvent } from "@sveltejs/kit";
 const resolveWithStatus = async <T extends RequestEvent>(
   status: number,
   statusText: string,
-  resolve: (event: T) => MaybePromise<Response>,
```
larsmaxwell, line 13: `MaybePromise` is no longer exported, so I took the type and manually entered it here.

vercel.json
```diff
@@ -2,5 +2,5 @@
   "buildCommand": "npm run build",
   "devCommand": "npm run dev",
   "framework": "sveltekit",
-  "installCommand": "npm install"
```
larsmaxwell, line 5: This is probably not necessary and can be restored

vite.config.ts
```diff
@@ -3,7 +3,6 @@ import { sveltekit } from "@sveltejs/kit/vite";
 import { partytownVite } from "@builder.io/partytown/utils";
 import svg from "@poppanator/sveltekit-svg";
 import type { PluginOption } from "vite";
-import { purgeCss } from "vite-plugin-svelte-purgecss";
```
larsmaxwell, line 6: @LouisFettet @hinzed1127 What happened when I upgraded the packages and corresponding dependencies was that `purgeCss` has changed to only support tailwindCSS. Should I try to replace this package?

larsmaxwell:
@LouisFettet @hinzed1127 After updating the dependency tree, the coverage has drastically decreased. Should I try to add some more ignore rules?

larsmaxwell:
Another issue that is popping up is that storybook cannot load icons from child components. I did a bit of debugging around this but a possible solution in the future would be to rewrite the way the storybook component is being loaded, or maybe even change the child component so that icons are loading independently of the nested component.
Screenshot 2024-10-10 at 4 42 38 PM

vite.config.ts
hinzed1127: that package was still in flux the last time I tried doing that, but it's probably worth another look. it looks like [the new plugin has a legacy mode](https://github.com/AdrianGonz97/vite-plugin-tailwind-purgecss/blob/master/legacy-mode.md) that targets beyond tailwind classes, with the caveat that it had more bugs. It's worth bringing it in and seeing if it hopefully still works the same way. If not, we can try to troubleshoot further

src/app.scss
hinzed1127: I would just note that in a comment here. Adds additional context if it becomes something to fix/return to in the future

vite.config.ts
larsmaxwell: @hinzed1127 Thanks for telling me about this. It looks like it works, but I have to put in an empty tailwind.config.js file 🤷

LouisFettet:
Superseded by #590.

(Thanks for your work here Larsen, it was really useful!)

PR #573, opened 2024-10-03
Add Recent News and Events Entity relationships to TopTier and ServiceEntry

### LINKED ISSUE #572: Adds Recent News to Core Content content type
DESCRIPTION TBA

-

## Acceptance criteria

- [ ]

larsmaxwell:
## Proposed changes
 - add entity relationship for recent news and events to core content page
 - add entity relationship for recent news and events to service entry page

## Screenshots

Screenshot 2024-10-04 at 12 12 54 PM

Top Tier:
Screenshot 2024-10-04 at 11 21 01 AM
Screenshot 2024-10-04 at 11 27 35 AM

- [x] Add Recent News and Events Entity relationships to TopTier and ServiceEntry

### Possible drawbacks

Currently, the tagged content overrides the entity relationships if the tagged content exists. This could be confusing for editors until all the tagged content is removed.

## Requested feedback

### DIFF
--- a/src/routes/(infoPages)/[topTierPage]/+page.server.ts
+++ b/src/routes/(infoPages)/[topTierPage]/+page.server.ts
@@ -12,11 +12,13 @@ import {
 } from "$lib/util/dates";
 
 import type { TaggedNewsAndEventsQuery, TopTierCollectionQuery } from "./$queries.generated";
+// assetProps are included with entryProps
 import entryPropsFragment from "$lib/fragments/entryProps";
 
 const query = gql`
-  ${entryPropsFragment}
+  # eslint-disable @graphql-eslint/selection-set-depth
 
+  ${entryPropsFragment}
   query TopTierCollection(
     $metadataID: String!
     $newsOldestDate: DateTime!
@@ -272,6 +274,7 @@ export const load = async ({
                 : undefined,
             }
           : undefined,
+        // relatedNews and relatedEvents default to tagged data
         relatedNews: taggedData?.newsCollection?.items.length
           ? taggedData?.newsCollection
           : matchedTopTier.recentNewsCollection?.items.length
--- a/src/routes/(infoPages)/[topTierPage]/+page.server.ts
+++ b/src/routes/(infoPages)/[topTierPage]/+page.server.ts
@@ -12,8 +12,11 @@ import {
 } from "$lib/util/dates";
 
 import type { TaggedNewsAndEventsQuery, TopTierCollectionQuery } from "./$queries.generated";
+import entryPropsFragment from '$lib/fragments/entryProps';
 
 const query = gql`
+  ${entryPropsFragment}
+
   query TopTierCollection(
     $metadataID: String!
     $newsOldestDate: DateTime!
@@ -97,26 +100,12 @@ const query = gql`
         }
         recentNewsCollection(where: { publicationDate_gt: $newsOldestDate }) {
           items {
-            sys {
-              id
-            }
-            type
-            title
-            subhead
-            publicationDate
-            slug
-            byline
+            ...EntryProps
           }
         }
         upcomingEventsCollection(where: { eventDateAndTime_gte: $eventStartDate }) {
           items {
-            sys {
-              id
-            }
-            slug
-            shortTitle
-            eventDescription
-            eventDateAndTime
+            ...EntryProps
           }
         }
       }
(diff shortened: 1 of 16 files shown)


src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.svelte
```diff
@@ -13,6 +13,8 @@
   import Image from "$lib/components/Image";
   import Link from "$lib/components/Link";
   import { url as arrowIcon } from "$icons/arrow_forward";
+  import NewsEntrySnippet from "../../about/news/NewsEntrySnippet.svelte";
+  import Event from "../../about/events/Event.svelte";
```
Comment (hinzed1127) [inline src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.svelte:17]: now that this is getting used across multiple pages, might make sense to move it under with `$lib/components` or another shared directory

src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts
```diff
@@ -151,6 +162,30 @@ const baseQuery = gql`
...
+          items {
+            sys {
+              id
+            }
+            slug
+            shortTitle
+            eventDescription
+            eventDateAndTime
+          }
+        }
```
Comment (hinzed1127) [inline src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts:188]: These _might_ be nice to have within `$lib/fragments/`. Mainly so that we're more aware of query chunks that are identical across different pages

hinzed1127 (approved):
Looks good, just a few thoughts on organizing some of the code that's reusable

src/routes/(infoPages)/[topTierPage]/+page.server.ts
```diff
@@ -90,6 +95,30 @@ const query = gql`
...
+          items {
+            sys {
+              id
+            }
+            slug
+            shortTitle
+            eventDescription
+            eventDateAndTime
+          }
+        }
```
Comment (LouisFettet) [inline src/routes/(infoPages)/[topTierPage]/+page.server.ts:121]: We should probably also set a `limit` here of some sort. I think 5 each of news and events should be sufficient.

src/routes/(infoPages)/[topTierPage]/+page.server.ts
```diff
@@ -254,8 +283,16 @@ export const load = async ({
...
+        relatedNews: taggedData?.newsCollection?.items.length
+          ? taggedData?.newsCollection
+          : matchedTopTier.recentNewsCollection?.items.length
+          ? matchedTopTier.recentNewsCollection
+          : { items: [] },
+        relatedEvents: taggedData?.eventEntryCollection?.items.length
+          ? taggedData?.eventEntryCollection
+          : matchedTopTier.upcomingEventsCollection?.items.length
+          ? matchedTopTier.upcomingEventsCollection
+          : { items: [] },
```
Comment (LouisFettet) [inline src/routes/(infoPages)/[topTierPage]/+page.server.ts:295]: LGTM, maybe just add a comment here that tagged news and events will get priority until we cutover to using the new fields.

src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts
Comment (LouisFettet) [reply src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.server.ts:188]: Yeah good call since we're using the same query here as on Top Tier. I forget if we can include variables (e.g. `$newsOldestDate` and `$eventStartDate`) in fragments, but we should at least be able to move the queried fields into a fragment.

src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.svelte
Comment (LouisFettet) [reply src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.svelte:17]: Yeah, that would definitely be ideal. I think at this point though a refactor would require a rather large lift, and it could spiral pretty quickly. I vote we keep it as-is for now to keep this PR simple, and we can tackle the refactor later (hopefully, if we have time).

src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.svelte
```diff
@@ -247,3 +251,21 @@
...
+{#if news?.items && news.items.length > 0}
+  <h2>Recent news</h2>
+  {#each news.items as entry (entry?.sys.id)}
+    {#if entry}
+      <NewsEntrySnippet {entry} headingLevel={4} />
+    {/if}
+  {/each}
+{/if}
+
+{#if events && events.items.length > 0}
```
Comment (LouisFettet) [inline src/routes/(infoPages)/[topTierPage]/[...serviceGroupPage]/+page.svelte:264]: I think we want this instead of consistency with `news` above, yeah?
```suggestion
{#if events?.items && events.items.length > 0}
```

LouisFettet (commented):
This is looking good! Adding a few of my thoughts as well.

PR #576, opened 2024-10-08
[Issue #575] Migrate to Vercel Speed Insights package.

### LINKED ISSUE #575: Migrate to Vercel Speed Insights package.
Related issues: N/A

## Problem statement

We're currently getting an alert on the [Speed Insights tab in Vercel](https://vercel.com/ldaf/ldaf/speed-insights) telling us that we need to update to using the new `@vercel/speed-insights` package.

## Background

We are currently sending stats to the [Speed Insights Intake API](https://vercel.com/docs/speed-insights/api), which is set up in the `src/lib/vitals.ts` file.

## Acceptance criteria

- [ ] We are using the new package to send data to Vercel and can confirm all metrics are working as intended.
- [ ] We are no longer sending data to the Speed Insights Intake API.

* Given there's a bit of customization in `src/lib/vitals`, it's unclear if we'll be able to duplicate our current solution with the new package. We should try to get as close as possible to the current functionality.
* If possible, we should defer any reporting to Vercel until the user has fully loaded the page (to cut down on bandwidth usage).
* More information from Vercel:
  * [Speed Insights Configuration with @vercel/speed-insights](https://vercel.com/docs/speed-insights/package)
  * [Migrating to the latest Speed Insights package](https://vercel.com/docs/speed-insights/migrating-from-legacy)

LouisFettet:
## Closes #575

The Speed Insights Intake API is being deprecated, so we need to switch to using the new `@vercel/speed-insights` package.

My first few attempts at this were focused on a number of different goals which all didn't work out as I expected:
- Only injecting the script in the production environment.
  - Turns out this is handled by the new package by default. It will load a `debug` script when running locally.
- Only injecting the script when the `layout` file is being run by the browser.
  - This is also handled by the new package; the SveleKit wrapper provided by the package makes the same check we were previously running for `web-vitals`.
- Setting up a way for us to easily debug events locally.
  - For the life of me I couldn't find a good way to do this. The `debug` mode that runs locally only ever prints a message indicating whether the script was loaded successfully. We could use the `beforeSend` method to do some logging, but this only runs in the `production` version of the script which isn't available when running locally (since it's hosted by Vercel on the domain).
- Sending events on every page load.
  - Turns out we were probably doing this incorrectly before. Since we're using client-side routing by default, the initial page load always takes longer than subsequent page loads (once you've loaded the first page on your visit, every page load after that is swapping out the content, not performing a full load). Since there doesn't seem to be a way to separate "initial load" events from "subsequent load" events, and since the package only runs on the initial page load by default, I gave up on trying to send events for subsequent page loads.
  
The result of all the above trial-and-error is that we're just doing the default implementation suggested in the docs: a very simple call to `injectSpeedInsights` in the base layout file.

## Proposed changes

- Remove references to `VERCEL_ANALYTICS_ID` since it's no longer needed.
- Remove the `web-vitals` package and add the `@vercel/speed-insights` package.
- Remove `src/lib/vitals.ts` and its invocation in `src/routes/+layout.svelte`, replacing it with the call to `injectSpeedInsights` from the new package.

- [x] We are using the new package to send data to Vercel and can confirm all metrics are working as intended.
- [x] We are no longer sending data to the Speed Insights Intake API.

## TODO

- [ ] After merging and deploying, we should remove the `PUBLIC_VERCEL_SPEED_INSIGHTS_ANALYTICS_ID` environment variable from the project in Vercel. It's unclear if this is even in-use currently, but it's best to deploy these changes first to be safe.


hinzed1127:
> Since there doesn't seem to be a way to separate "initial load" events from "subsequent load" events, and since the package only runs on the initial page load by default, I gave up on trying to send events for subsequent page loads.

They give [the same caveat](https://vercel.com/docs/speed-insights/metrics#understanding-data-points) with Next.js-based sites:

> Data points are collected on hard navigations, which in the case of Next.js apps, are only the first-page view in a session. During a user's visit, data points are gathered during the initial page load, user interaction, and upon leaving the page. 

So it seems like that's a known constraint and something we can roll with for now.

hinzed1127 (approved):

LouisFettet:
Of course the one page of docs I didn't read was the one that answered my open questions around initial vs subsequent page load 🤦‍♂️ 

Thanks for linking to that @hinzed1127, I feel much better about the behavior here now!

PR #579, opened 2024-10-16
News and Events Search

### LINKED ISSUE #578: Index News and Events in site search
Related issues:

-

## Acceptance criteria

- [ ]

larsmaxwell:
## Proposed changes

- Adds separate utility for mapping `eventEntry` and `news` content types under `/src/lib/loadEventsNewsMetadataMap.ts`. 
- Groups events and news entity update under the api/v1/search-index/ endpoint
- Adds additional `GET` endpoint for dumping `pageMetadata`, `news`, and `eventEntry` json into Algolia

## Screenshots
Screenshot 2024-10-18 at 8 45 03 AM
Screenshot 2024-10-18 at 8 47 07 AM

Was able to test locally using https://webhook.site/ and postman


src/lib/services/server/contentful/schema.graphql
```diff
@@ -2968,7 +3028,6 @@ type PageMetadata implements Entry & _Node {
   metaDescription(locale: String): String
   metaTitle(locale: String): String
   parent(locale: String, preview: Boolean, where: PageMetadataFilter): PageMetadata
-  recentNewsCollection(limit: Int = 100, locale: String, preview: Boolean, skip: Int = 0): PageMetadataRecentNewsCollection
```
LouisFettet, line 2971: This field was the one we accidentally added to `Page details`, right? In other words, I assume that this deletion is intentional since we added this field to the Top Tier and Core Content pages instead?

src/lib/types.ts
```diff
@@ -1,5 +1,46 @@
+import type { Breadcrumbs } from "$lib/components/Breadcrumbs";
+
 export type User = {
   email: string;
   name: string;
   avatarURL: string;
 };
+
+export type MetadataMapItem = {
```
LouisFettet, line 9: I've had a think about this, and while it would be nice to have a single type definition for this, I think this might be too generic. It's also divorced from the Contentful schema, which could make it brittle.

Instead of trying to reuse the existing `PageMetadataMapItem` type in `src/lib/loadEventsNewsMetadataMap.ts`, we should focus on fixing whatever is wrong with the type definition created by the `gen-operations` script that's getting output to `src/lib/$queries.generated.d.ts`.

src/lib/util/dates.ts
```diff
@@ -10,6 +10,12 @@ export const getCurrentDateInTZ = (tz: string): string => {
   return formatISO(zoned, { format: "extended", representation: "date" });
 };
 
+export const getDateTwelveMonthsAgoInTZ = (tz: string): string => {
+  const date = sub(new Date(), { months: 12 });
+  const zoned = utcToZonedTime(date, tz);
+  return formatISO(zoned, { format: "extended", representation: "date" });
+};
+
```
LouisFettet, line 18: Is this being used anywhere? We should probably remove it if not.

src/lib/loadEventsNewsMetadataMap.ts
```diff
@@ -0,0 +1,100 @@
...
+      slug?: string | null;
+      eventDateAndTime?: string | null;
+      eventDescription?: string | null;
+    }[];
+  };
+};
+
+// TODO: remove
+const constructEventSlug = (date: Date, slug: string): string =>
+  `${date.toISOString().split("T")[0]}-${slug}`;
```
LouisFettet, line 62: We should use the [new util](https://github.com/la-ldaf/ldaf-site/blob/b1b3602037af48cfa8291a7667f036aefcf4df27/src/lib/util/constructEventSlug.ts) instead.

src/routes/api/v1/search-index/+server.ts
```diff
@@ -15,36 +17,20 @@ const CONTENTFUL_ACTIONS = {
 };
 
 export const POST = async ({ request, locals: { contentfulClient } }) => {
-  authenticateRequest(request);
+  // authenticateRequest(request);
```
LouisFettet, line 20: We'll want to uncomment this before merge, yeah?

src/routes/api/v1/search-index/get-search-index.json
LouisFettet, line 1: This file should be removed as well, yeah?

LouisFettet (commented):
Great work here @larsmaxwell! The only major thing I think we need to look into is the type definitions. Otherwise, most of my comments are about cleaning up some things. If I have time I might take a crack at getting the correct defs generated so that we don't need to hardcode anything that we should be able to infer from the Contentful schema.

I'll leave final approval here up to @hinzed1127 since he's got a lot more expertise with  Algolia.

src/lib/types.ts
larsmaxwell: I think the problem that I was trying to solve here was that in order to reuse some code, the two types had some differences that did not overlap, so I wanted to create a type that was more flexible to support metadata map item of page, event, and news.

larsmaxwell: I'll look into fixing this, but I may need some help with type generation.

src/lib/services/server/contentful/schema.graphql
LouisFettet: Cool, thanks for the gut-check!

src/lib/types.ts
LouisFettet: Posted this in Slack, but I think #583 should resolve some of the problems here (and correct the query).

LouisFettet (commented):
I addressed most of my major concerns in the PR that @larsmaxwell already merged into this one. The remaining items here that I think we should look into are:
1. Resolving the merge conflicts with the schema; this should be as simple as switching `CONTENTFUL_DEFAULT_ENVIRONMENT` to `master` and running `npm gen-schema`.
2. Figuring out what to do about the `authenticateRequest` function on the indexing endpoint.
3. Removing the static json files, as I assume those were only for testing purposes.

src/routes/api/v1/search-index/+server.ts
hinzed1127: we will, I'm guessing this was just due to @larsmaxwell getting around not having this set up for local development

src/lib/types.ts
LouisFettet, line 1: Ah thanks, I thought I had removed these changes but I must have missed them.

src/routes/api/v1/search-index/+server.ts
```diff
@@ -14,37 +17,39 @@ const CONTENTFUL_ACTIONS = {
   DELETE: "ContentManagement.Entry.delete",
 };
 
+const getMetaTitleAndDescription = (
+  item?: SearchIndexingMetadataMapItem,
+): { metaTitle?: string | null; metaDescription?: string | null } => {
+  let metaTitle, metaDescription;
+  if (item?.__typename == "PageMetadata") {
+    metaTitle = item?.metaTitle ?? item?.title;
+    metaDescription = item?.metaDescription;
```
hinzed1127, line 26: we're gonna have to fix this. There reason for the `for fields in body.fields` loop [below](https://github.com/la-ldaf/ldaf-site/pull/579/files#diff-18e3587586f7ba5303d44f50cf6bffdebc3fb35c13b16bd4d4ccc000abf1f8aaR65) is because Contentful's webhook only sends fields that have changed in the request body, so we end up skipping any properties that aren't sent in the request body. Since this explicitly checks for the title and description, this could unintentionally set fields to undefined when they haven't changed.

src/routes/api/v1/search-index/+server.ts
```diff
@@ -100,3 +107,29 @@ export const POST = async ({ request, locals: { contentfulClient } }) => {
     throw error(400, message as string);
   }
 };
+
+// Return all suitable Page Metadata, News, and Events map items
+export const GET = async ({ locals: { contentfulClient } }) => {
```
hinzed1127, line 112: `TODO` for my service entry updates: add a query param to quickly filter between `metadata`, `news-and-events`, and (from mine) `service-entries`

hinzed1127 (approved):
This has a few issues that we discussed (and I also commented on here) around unintentionally overwriting title and description values if they aren't present in the request body. I'd rather address it in https://github.com/la-ldaf/ldaf-site/pull/582/. The easier way might be to simply pull the news and events updates into their own endpoint like service entries. We'll see how feasible that ends up being

hinzed1127:
2 other things:
1. Looks like you need to run prettier to fix linting issues first
2. Separately, we'll manually add news and events to algolia to initially populate the index

PR #583, opened 2024-10-31
[Issue #578] Iimprove type definitions for search indexing.

LouisFettet:
Keeping the description bare here; this is an attempt to fix up type definitions by utilizing the generated definitions and performing some eldritch TS magic.

### DIFF
--- a/src/lib/components/ContentfulRichText/nodes/EntryHyperlink.svelte
+++ b/src/lib/components/ContentfulRichText/nodes/EntryHyperlink.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
   import type { Node as NodeType, EntryHyperlink } from "@contentful/rich-text-types";
-  import type { MetadataMapItem } from "$lib/types";
+  import type { PageMetadataMapItem } from "$lib/loadPageMetadataMap";
   import Node from "./Node.svelte";
   import { getContext } from "svelte";
   import { linksKey, type LinksContext } from "../context";
@@ -21,7 +21,7 @@
 
   const entry = linksContext.linksEntriesMaps.hyperlink.get(entryID);
 
-  let entryMetadata: MetadataMapItem;
+  let entryMetadata: PageMetadataMapItem;
   if (entry?.__typename === "PageMetadata") {
     linksContext.pageMetadataMap.get(entryID);
     const meta = linksContext.pageMetadataMap.get(entryID);
--- a/src/lib/components/SideNav/SideNav.server.ts
+++ b/src/lib/components/SideNav/SideNav.server.ts
@@ -1,5 +1,4 @@
-import type { PageMetadataMap } from "$lib/loadPageMetadataMap";
-import type { MetadataMapItem } from "$lib/types";
+import type { PageMetadataMap, PageMetadataMapItem } from "$lib/loadPageMetadataMap";
 import type { NavMenuType } from "$lib/components/Header/Nav/types";
 import type SideNavItem from "./types";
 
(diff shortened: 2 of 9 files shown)


src/lib/components/ContentfulRichText/nodes/EntryHyperlink.svelte
LouisFettet, line 1: Reverted without changes.

src/lib/components/SideNav/SideNav.server.ts
LouisFettet, line 1: Reverted without changes.

src/lib/loadEventsNewsMetadataMap.ts
```diff
@@ -1,10 +1,13 @@
...
+import constructEventSlug from "./util/constructEventSlug";
 import type { ContentfulClient } from "$lib/services/server/contentful";
-import type { MetadataMapItem } from "./types";
+import type { PageMetadataMapItem } from "$lib/loadPageMetadataMap";
+
+import type { NewsAndEventsCollectionQuery } from "./$queries.generated";
 
 const queryNewsAndEvents = gql`
-  query NewsCollection {
+  query NewsAndEventsCollection {
```
LouisFettet, line 10: We were missing the new meta fields below in the query. We should also be using `eventSummary` over `eventDescription`.

src/lib/loadEventsNewsMetadataMap.ts
```diff
@@ -25,48 +30,35 @@ const queryNewsAndEvents = gql`
...
+
+type AdditionalMetadataFields = { url?: string | null; children?: string[] };
 
-// TODO: remove
-const constructEventSlug = (date: Date, slug: string): string =>
-  `${date.toISOString().split("T")[0]}-${slug}`;
+export type SearchIndexingMetadataMapItem =
+  | PageMetadataMapItem
+  | (NewsSearchIndexingMetadataMapItem & AdditionalMetadataFields)
+  | (EventSearchIndexingMetadataMapItem & AdditionalMetadataFields);
```
LouisFettet, line 54: This isn't perfect, but it does give us what we need to ensure we're not querying for fields that don't exist. The downside here is that we need to check the `__typename` for the fields that don't overlap, but I think that's a fair trade-off and makes it easier to reason about.

src/lib/loadPageMetadataMap.ts
LouisFettet, line 1: This file was reverted, with the only change being that I moved the `PageMetadataMapItemWithObjectID` definition into the Algolia status API route.

src/routes/+layout.svelte
LouisFettet, line 1: Reverted without changes.

src/routes/+page.server.ts
LouisFettet, line 1: Reverted without changes.

src/routes/api/v1/search-index/+server.ts
```diff
@@ -16,6 +17,23 @@ const CONTENTFUL_ACTIONS = {
...
+    metaDescription = item?.metaDescription;
+  } else if (item?.__typename === "News") {
+    metaTitle = item?.metaTitle ?? item?.title;
+    metaDescription = item?.metaDescription ?? item?.subhead;
+  } else if (item?.__typename === "EventEntry") {
+    metaTitle = item?.metaTitle ?? item?.shortTitle;
+    metaDescription = item?.metaDescription ?? item?.eventSummary;
+  }
+  return { metaTitle, metaDescription };
+};
```
LouisFettet, line 35: This is a bit ugly, but we can be much more certain of the data we're looking at by switching on the `__typename`.

LouisFettet (commented):
Self-review!

larsmaxwell (approved):

PR #585, opened 2024-11-01
Fire Weather Map Updates

hinzed1127:
## Proposed changes

- Integrates Contentful with the Fire Weather Map
   - Added a [single-use content type](https://app.contentful.com/spaces/pc5e1rlgfrov/content_types/fireDangerMap/fields) for that page.
   - [Content content using that content type](https://app.contentful.com/spaces/pc5e1rlgfrov/entries/7fiQCwkywcWi7lBy0dayqd?focusedField=title)

## Screenshots

Notice that it gets included in the navigation side bar:
Screenshot 2024-10-31 at 9 49 10 PM

Description rich text served from Contentful:
Screenshot 2024-10-31 at 9 49 20 PM

- [x] Map route exists as page details in Contentful
- [x] Description for the map lives in and is served from Contentful

### Alternate solutions

- In theory, we could use a core content page, since this is a subset of those properties, but I think it's better to explicitly restrict it to just the description alongside the map.

### DIFF
--- a/src/routes/(infoPages)/land/fire/safety/fire-danger-map/+page.server.ts
+++ b/src/routes/(infoPages)/land/fire/safety/fire-danger-map/+page.server.ts
@@ -1,6 +1,22 @@
+import gql from "graphql-tag";
+import { print as printQuery } from "graphql";
 import parishTopoJSON from "./parishes.json";
 
-export const load = async ({ locals: { getKVClient } }) => {
+import type { FireWeatherMapQuery } from "./$queries.generated";
+
+const query = gql`
+  query FireWeatherMap {
+    fireDangerMapCollection(limit: 1) {
+      items {
+        description {
+          json
+        }
+      }
+    }
+  }
+`;
+
+export const load = async ({ locals: { getKVClient }, locals: { contentfulClient } }) => {
   try {
     const kvClient = await getKVClient();
     const fireWeatherData = await kvClient.getFireWeatherData();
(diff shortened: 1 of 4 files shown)


src/routes/(infoPages)/land/fire/safety/fire-danger-map/+page.server.ts
```diff
@@ -1,6 +1,22 @@
...
+  query FireWeatherMap {
+    fireDangerMapCollection(limit: 1) {
+      items {
+        description {
+          json
+        }
+      }
+    }
+  }
+`;
```
LouisFettet, line 17: This could have an adverse side-effect if there's more than one entry erroneously created; I'd recommend doing something like we did on `/licensing-permits` where we look up the  pageMetadata ID based on the route (and then query for the one entry instead of a collection): https://github.com/la-ldaf/ldaf-site/blob/9aa03aa9241b5f32bad23bbb8245454944269393/src/routes/licensing-permits/%2Bpage.server.ts#L95

src/routes/(infoPages)/land/fire/safety/fire-danger-map/+page.svelte
```diff
@@ -92,7 +98,27 @@
...
+                  } else {
+                    let formattedDate;
+                    try {
+                      formattedDate = format(
+                        new Date(d.properties.ObservationDate),
+                        "MMMM d, yyyy",
+                      );
+
+                      return `Parish: ${parishName}\nFire Risk: ${
+                        fireRiskLevels[Number(fireRisk)]
```
LouisFettet, line 114: I think we have an off-by-one error here; I'm seeing the following on the preview:
* green: medium
* yellow: very high
* orange: extreme
* red: undefined

LouisFettet (commented):
Noticed two issues that I've noted in code review that we should look into before merging.

Additionally, the page heading is an `h2`, but it should be an `h1`. I'm unsure where that's getting set (I assume it's by the Plot?).

LouisFettet (approved):
Pre-approving so I don't block you, but yeah I think at the least we need to handle the issue with the heading and the labels that appear on the parishes before merging.

src/routes/(infoPages)/land/fire/safety/fire-danger-map/+page.svelte
hinzed1127: ah, yup. forgot about the array index offset. easy fix.

PR #586, opened 2024-11-01
[No Issue] Improve fetching of content and metadata for the fire weather map page.

LouisFettet:
Decided to make this a separate PR to talk through some of these changes in a self-review.

### DIFF
--- a/src/routes/(infoPages)/land/fire/safety/fire-danger-map/+page.server.ts
+++ b/src/routes/(infoPages)/land/fire/safety/fire-danger-map/+page.server.ts
@@ -1,47 +1,69 @@
 import gql from "graphql-tag";
 import { print as printQuery } from "graphql";
+import { error } from "@sveltejs/kit";
 import parishTopoJSON from "./parishes.json";
 
-import type { FireWeatherMapQuery } from "./$queries.generated";
+import type { FireWeatherMapPageQuery } from "./$queries.generated";
 
 const query = gql`
-  query FireWeatherMap {
-    fireDangerMapCollection(limit: 1) {
+  query FireWeatherMapPage($metadataID: String!, $preview: Boolean = false) {
+    fireDangerMapCollection(
+      where: { pageMetadata: { sys: { id: $metadataID } } }
+      limit: 1
+      preview: $preview
+    ) {
       items {
+        title
         description {
           json
+        }
+        pageMetadata {
+          sys {
+            id
+          }
         }
       }
     }
   }
 `;
 
-export const load = async ({ locals: { getKVClient }, locals: { contentfulClient } }) => {
-  try {
-    const kvClient = await getKVClient();
-    const fireWeatherData = await kvClient.getFireWeatherData();
+export const load = async ({ parent, locals: { getKVClient, contentfulClient } }) => {
+  fetchData: {
+    if (!contentfulClient) break fetchData;
+    const { pageMetadataMap, pathsToIDs } = await parent();
+    const metadataID = pathsToIDs.get("/land/fire/safety/fire-danger-map");
+    if (!metadataID) break fetchData;
+    const pageMetadata = pageMetadataMap.get(metadataID);
+    if (!pageMetadata) break fetchData;
 
-    if (fireWeatherData === undefined) {
-      throw new Error("Error fetching fire weather data.");
+    const pageData = await contentfulClient.fetch<FireWeatherMapPageQuery>(printQuery(query), {
+      variables: { metadataID },
+    });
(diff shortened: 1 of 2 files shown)


src/routes/(infoPages)/land/fire/safety/fire-danger-map/+page.server.ts
```diff
@@ -1,47 +1,69 @@
...
       items {
+        title
         description {
           json
         }
+        pageMetadata {
+          sys {
+            id
+          }
+        }
```
LouisFettet, line 24: Besides making sure we grab the appropriate entry, I also added the `title` so that we could use whatever the editors provide in Contentful.

src/routes/(infoPages)/land/fire/safety/fire-danger-map/+page.server.ts
```diff
@@ -1,47 +1,69 @@
...
-      throw new Error("Error fetching fire weather data.");
-    }
+export const load = async ({ parent, locals: { getKVClient, contentfulClient } }) => {
+  fetchData: {
+    if (!contentfulClient) break fetchData;
+    const { pageMetadataMap, pathsToIDs } = await parent();
+    const metadataID = pathsToIDs.get("/land/fire/safety/fire-danger-map");
+    if (!metadataID) break fetchData;
+    const pageMetadata = pageMetadataMap.get(metadataID);
+    if (!pageMetadata) break fetchData;
```
LouisFettet, line 37: Added in the good ol' `fetchData` label so that we can more easily return a 404 in the case of content errors (as opposed to map data errors).

The other benefit here is that this gives us the `pageMetadata` entry so that we can update the meta tags and such ([in the base layout](https://github.com/la-ldaf/ldaf-site/blob/9aa03aa9241b5f32bad23bbb8245454944269393/src/routes/%2Blayout.svelte#L55)).

src/routes/(infoPages)/land/fire/safety/fire-danger-map/+page.server.ts
```diff
@@ -1,47 +1,69 @@
...
+          topoJSON: parishTopoJSON,
+        },
+        pageData: fireDangerMapPageData,
+      };
+    } catch (error) {
+      const errorMessage = error instanceof Error ? error.message : "An unknown error occurred";
+      return {
+        error: errorMessage,
+      };
+    }
```
LouisFettet, line 66: This block is effectively unchanged, just nested in `fetchData` with the content fetching moved further up.

src/routes/(infoPages)/land/fire/safety/fire-danger-map/+page.svelte
```diff
@@ -8,7 +8,6 @@
 
   export let data;
   $: ({ fireWeatherData, pageData, error = null } = data);
-  $: description = pageData?.description;
```
LouisFettet, line 11: I think since `pageData` is already reactive, we don't need `description` (or `title`) to also be reactive. Unless I'm missing something here?

LouisFettet (commented):
self-review!

src/routes/(infoPages)/land/fire/safety/fire-danger-map/+page.svelte
hinzed1127: You're probably right. My memory of some of the reactivity stuff in Svelte was foggy, so I just defaulted to the greedy case, but I think that's unnecessary

src/routes/(infoPages)/land/fire/safety/fire-danger-map/+page.server.ts
hinzed1127: good catch, I forgot about some of those other tie-ins to `pageMetadata`

hinzed1127 (approved):
Looks good! Thank you for the assist on some tweaks/cleanup

src/routes/(infoPages)/land/fire/safety/fire-danger-map/+page.server.ts
LouisFettet: Ah okay, yeah I didn't see that help text (but I also might have missed it) so I figured it would be more confusing for the title to not be in-use.

src/routes/(infoPages)/land/fire/safety/fire-danger-map/+page.svelte
LouisFettet: nb4 [SvelteKit 5's runes](https://vercel.com/blog/whats-new-in-svelte-5#runes) drastically change how all of this works 😵

PR #587, opened 2024-11-01
Refactor news and events

hinzed1127:
There's enough here that I wanted to separate it out into a side-PR before merging everything.

On a high level, this: 

- breaks out news and events updates into their own API endpoint and contentful webhook
- reverts some of the mapping in the original algolia endpoint to make sure `pageMetadata` records don't get accidentally undefined

### DIFF
--- a/src/routes/api/v1/search-index/+server.ts
+++ b/src/routes/api/v1/search-index/+server.ts
@@ -17,39 +17,23 @@ const CONTENTFUL_ACTIONS = {
   DELETE: "ContentManagement.Entry.delete",
 };
 
-const getMetaTitleAndDescription = (
-  item?: SearchIndexingMetadataMapItem,
-): { metaTitle?: string | null; metaDescription?: string | null } => {
-  let metaTitle, metaDescription;
-  if (item?.__typename == "PageMetadata") {
-    metaTitle = item?.metaTitle ?? item?.title;
-    metaDescription = item?.metaDescription;
-  } else if (item?.__typename === "News") {
-    metaTitle = item?.metaTitle ?? item?.title;
-    metaDescription = item?.metaDescription ?? item?.subhead;
-  } else if (item?.__typename === "EventEntry") {
-    metaTitle = item?.metaTitle ?? item?.shortTitle;
-    metaDescription = item?.metaDescription ?? item?.eventSummary;
-  }
-  return { metaTitle, ...(metaDescription ? { metaDescription } : {}) };
-};
-
 export const POST = async ({ request, locals: { contentfulClient } }) => {
   authenticateRequest(request);
 
   const contentfulAction = request.headers.get("x-contentful-topic") || "";
   const body = await request.json();
   const contentType = body?.sys?.contentType?.sys?.id;
-  const contentTypes = ["pageMetadata", "news", "eventEntry"];
+  // TODO: in the future, we might want to consolidate these search index endpoints
+  // to handle a wider range of contentTypes in a single POST endpoint. As it stands,
+  // the transform logic is varied enough that it's easier to have some duplicative scaffolding
+  // for calling algolia in a similar way but with different transformation logic in the
+  // "Contentful webhook -> SvelteKit API route -> Algolia index" pipeline
+  const contentTypes = ["pageMetadata"];
   const { pageMetadataMap } = await loadPageMetadataMap({ contentfulClient });
-  const eventsAndNewsMap = await loadEventsAndNewsMap({ contentfulClient });
-
-  const metadataMap = new Map([...eventsAndNewsMap, ...pageMetadataMap]);
 
   try {
     if (contentfulAction === CONTENTFUL_ACTIONS.PUBLISH && contentTypes.includes(contentType)) {
-      const contentfulValue = metadataMap.get(body.sys.id);
-      const metaTitleAndDescription = getMetaTitleAndDescription(contentfulValue);
+      const contentfulValue = pageMetadataMap.get(body.sys.id) || { url: "", children: [] };
       // The webhook body is missing `children` and `url`, since we
       // construct those in `loadPageMetadataMap`. Add those properties here.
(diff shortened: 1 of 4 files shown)


src/routes/api/v1/search-index/+server.ts
```diff
@@ -17,39 +17,23 @@ const CONTENTFUL_ACTIONS = {
...
-};
-
 export const POST = async ({ request, locals: { contentfulClient } }) => {
   authenticateRequest(request);
 
   const contentfulAction = request.headers.get("x-contentful-topic") || "";
   const body = await request.json();
   const contentType = body?.sys?.contentType?.sys?.id;
-  const contentTypes = ["pageMetadata", "news", "eventEntry"];
+  // TODO: in the future, we might want to consolidate these search index endpoints
```
hinzed1127, line 26: I basically reverted this endpoint into its original functionality of just handling pageMetadata updates. [I hardened that requirement in the webhook](https://app.contentful.com/spaces/pc5e1rlgfrov/settings/webhooks/0UGOXuEe2TiUSD8dCJjGpS/settings), too, so the original webhook is only getting triggered for PageMetadata Updates

src/routes/api/v1/search-index/+server.ts
```diff
@@ -17,39 +17,23 @@ const CONTENTFUL_ACTIONS = {
...
   const { pageMetadataMap } = await loadPageMetadataMap({ contentfulClient });
-  const eventsAndNewsMap = await loadEventsAndNewsMap({ contentfulClient });
-
-  const metadataMap = new Map([...eventsAndNewsMap, ...pageMetadataMap]);
 
   try {
     if (contentfulAction === CONTENTFUL_ACTIONS.PUBLISH && contentTypes.includes(contentType)) {
-      const contentfulValue = metadataMap.get(body.sys.id);
-      const metaTitleAndDescription = getMetaTitleAndDescription(contentfulValue);
+      const contentfulValue = pageMetadataMap.get(body.sys.id) || { url: "", children: [] };
```
hinzed1127, line 36: this is just a reversion to the previous state

src/routes/api/v1/search-index/+server.ts
```diff
@@ -58,9 +42,7 @@ export const POST = async ({ request, locals: { contentfulClient } }) => {
           id: body.sys.id,
         },
         url: contentfulValue?.url,
-        ...metaTitleAndDescription,
-        // conditionally add page attributes
-        ...{ ...(contentfulValue?.children ? { children: contentfulValue?.children } : {}) },
+        children: contentfulValue?.children,
```
hinzed1127, line 45: this is just a reversion to the previous state

src/routes/api/v1/search-index/+server.ts
```diff
@@ -108,6 +90,23 @@ export const POST = async ({ request, locals: { contentfulClient } }) => {
...
+  } else if (item?.__typename === "News") {
+    metaTitle = item?.metaTitle ?? item?.title;
+    metaDescription = item?.metaDescription ?? item?.subhead;
+  } else if (item?.__typename === "EventEntry") {
+    metaTitle = item?.metaTitle ?? item?.shortTitle;
+    metaDescription = item?.metaDescription ?? item?.eventSummary;
+  }
+  return { metaTitle, ...(metaDescription ? { metaDescription } : {}) };
+};
+
```
hinzed1127, line 109: this is only getting used by the GET endpoint below that @larsmaxwell added. I _did_ add a query filter to that endpoint, so if you wanna quickly view just one of the maps, you can do `?filter=pages` or `?filter=news-and-events`. Incomplete but something worth returning to in the future if needed

src/routes/api/v1/search-index/news-and-events/+server.ts
```diff
@@ -0,0 +1,115 @@
...
+    if (eventSummary) {
+      algoliaIndexObject.metaDescription = eventSummary;
+    } else if (eventDateAndTime) {
+      const date = parseISO(eventDateAndTime);
+      algoliaIndexObject.metaDescription = `Event on ${format(date, "MMMM d, yyyy")}`;
+    }
+  }
+};
+
+export const POST = async ({ request }) => {
```
hinzed1127, line 53: This is getting called by a [new webhook](https://app.contentful.com/spaces/pc5e1rlgfrov/settings/webhooks/0W7ypAtUzQICLr94Vl5cZg/settings) that is just triggered for news and events

src/routes/api/v1/search-index/news-and-events/+server.ts
```diff
@@ -0,0 +1,115 @@
...
+  if (algoliaIndexObject.metaDescription === undefined) {
+    const { eventSummary, eventDateAndTime } = algoliaIndexObject;
+    if (eventSummary) {
+      algoliaIndexObject.metaDescription = eventSummary;
+    } else if (eventDateAndTime) {
+      const date = parseISO(eventDateAndTime);
+      algoliaIndexObject.metaDescription = `Event on ${format(date, "MMMM d, yyyy")}`;
+    }
+  }
+};
```
hinzed1127, line 51: this is basically fleshing out what `getMetaTitleAndDescription`, but without the side effects that was having on `pageMetadata` entries

hinzed1127 (commented):
Self-review

src/routes/api/v1/search-index/+server.ts
larsmaxwell: If you're looking to populate the search index, it would be good to have the GET endpoints loading separate content types as well (one for page content/ one for news/events).

larsmaxwell (approved):

src/routes/api/v1/search-index/+server.ts
hinzed1127: Since the service entries are already taken care of, and there was only one event to add, this wasn't really needed for a one-time bulk import. But I think it's nice to have around from a utility standpoint in the future

